PR #2, opened 2022-07-17
Var reduction

valeriaRaffuzzi:
Hello, I think the basics here are done…

I expanded the VectorField to a WeightWindowsField, which has a TallyMap and some vectors to store the minimum and maximum weights. The weights are read from an external file, whose path is included in the input file.

The input is in the form:

varianceReduction { type weightWindowsField; file /home/vr339/weightWindows.txt ; }

which is read by the physicsPackage and given to the GeometryReg to initialise the field globally. The weight vectors in the WW input file need to have the same ordering as SCONE’s map bins.

Then the actual weights are retrieved in the implicit collision processor, in the function cut-off. I still wanted to maintain the possibility to do splitting and rouletting without weight windows, but if one activates both options by mistake, weight windows should have priority. For this reason I had to make a flag in collisionOperator for weight windows: if one wants to use them, they have to include both the WW file in the input file and the WW flag in the collision processor.

NOTE: It looks like there are conflicts now, that were not there when I created the PR. I'll fix them asap. I know we said we would work with forks, but with this old PR it's the fastest this way!


Mikolaj-A-Kowalski:
For the conflicts please try to rebase on the main branch 
e.g. in your branch `git rebase -i main` It will put your commits on top of the current `main` and allow you to resolve the conflicts.

CollisionOperator/CollisionProcessors/neutronCEimp_class.f90
```diff
@@ -162,13 +174,21 @@ subroutine init(self, dict)
...
     if (self % implicitAbsorption) then
       if (.not.self % roulette) call fatalError(Here,&
          'Must use Russian roulette when using implicit absorption')
       if (.not.self % implicitSites) call fatalError(Here,&
          'Must generate fission sites implicitly when using implicit absorption')
     end if
 
+    ! Sets up the weight windows field
+    if (self % weightWindows) then
+      name = 'WeightWindows'
```
Mikolaj-A-Kowalski, line 187: Is this a global constant name? 
Perhaps we should have it as a global constant if multiple parts of the code will be trying to use it.

CollisionOperator/CollisionProcessors/neutronCEimp_class.f90
```diff
@@ -452,43 +472,73 @@ subroutine cutoffs(self, p, collDat, thisCycle, nextCycle)
...
     ! This value must be at least 2
-    mult = ceiling(p % w/self % maxWgt)
+    n = floor(p % w/maxWgt)
+    prob = p % w/maxWgt - n
+
+    if (p % pRNG % get() < prob) then
+      mult = n + 1
+    else
+      mult = n
+    end if
```
Mikolaj-A-Kowalski, line 538: What about something along the lines of 
```Fortran
n = floor((p % w / maxWgt) + p % pRNG & get())
```
Which is how the probabilistic rounding is usually handled in case of fission site generation.

CollisionOperator/CollisionProcessors/neutronCEimp_class.f90
```diff
@@ -452,43 +472,73 @@ subroutine cutoffs(self, p, collDat, thisCycle, nextCycle)
...
     class(neutronCEimp), intent(inout)    :: self
     class(particle), intent(inout)        :: p
     class(particleDungeon), intent(inout) :: thisCycle
-    integer(shortInt)                     :: mult,i
+    real(defReal), intent(in)             :: maxWgt
+    integer(shortInt)                     :: mult, n, i
+    real(defReal)                         :: prob
 
     ! This value must be at least 2
-    mult = ceiling(p % w/self % maxWgt)
```
Mikolaj-A-Kowalski, line 491: What was the old "fractional" split not working in some way? 
Sorry, I am just no sure why it had too go :-) But I may be missing something obvious.

Geometry/Fields/VectorFields/Tests/weightWindowsField_test.f90
```diff
@@ -0,0 +1,87 @@
...
+    procedure :: setUp
+    procedure :: tearDown
+  end type test_weightWindows
+
+
+  !!
+  !! Weight Windows Definition
+  !!
+  character(*), parameter :: DICT_DEF = &
+  & "file /home/vr339/SCONE/scone/IntegrationTestFiles/testWW ;"
```
Mikolaj-A-Kowalski, line 26: If it is using a file it needs to become an 'Integration Test' (with `_iTest` suffix). Also path should be relative to the git root directory i.e. `IntegrationTestFiles/testWW ;`

Geometry/Fields/VectorFields/weightWindowsField_class.f90
```diff
@@ -0,0 +1,165 @@
...
+  !!
+  public :: weightWindowsField_TptrCast
+
+  !!
+  !! Weight Windows Vector Field
+  !!
+  !! Returns a vector with lower, upper and survival weight
+  !!
+  !! Sample Dictionary Input:
+  !!   varianceReduction { type weightWindowsField; file /home/WWfile.txt ;}
```
Mikolaj-A-Kowalski, line 29: Perhaps we should include some description of the expected contents of the file in the doc-comment?

Geometry/Fields/VectorFields/weightWindowsField_class.f90
```diff
@@ -0,0 +1,165 @@
...
+    call new_tallyMap(self % net, dict2 % getDictPtr('map'))
+    self % N = self % net % bins(ALL)
+
+    ! Read weight values from file
+    call dict2 % get(self % constSurvival, 'constSurvival')
+    call dict2 % get(self % lowerW, 'wLower')
+    call dict2 % get(self % upperW, 'wUpper')
+
+    ! Check data consistency
+    if (size(self % lowerW) /= self % N) then
```
Mikolaj-A-Kowalski, line 86: `upperW`  probably also needs this check, right?

Geometry/Fields/VectorFields/weightWindowsField_class.f90
```diff
@@ -0,0 +1,165 @@
...
+  end subroutine init
+
+  !!
+  !! Return to uninitialised state
+  !!
+  elemental subroutine kill(self)
+    class(weightWindowsField), intent(inout) :: self
+
+    call self % net % kill()
+    deallocate(self % net)
```
Mikolaj-A-Kowalski, line 106:
```suggestion
    if (allocated(self % net)) deallocate(self % net)
    self % N = 0
    self % constSurvival = ZERO
```

Geometry/Fields/VectorFields/weightWindowsField_class.f90
```diff
@@ -0,0 +1,165 @@
...
+    integer(shortInt)                     :: binIdx
+
+    ! Get current particle state
+    state = p
+
+    ! Read map bin index
+    binIdx = self % net % map(state)
+
+    ! Return if invalid bin index
+    if (binIdx == 0) then
```
Mikolaj-A-Kowalski, line 129: (Note to self) How happy are different parts of the code to receive vector of 0s as weight window parameters.

CollisionOperator/CollisionProcessors/neutronCEimp_class.f90
```diff
@@ -452,43 +472,73 @@ subroutine cutoffs(self, p, collDat, thisCycle, nextCycle)
...
 
-    if (p % E < self % minE ) then
+    if (p % E < self % minE) then
       p % isDead = .true.
+    ! Weight Windows treatment
+    elseif (self % weightWindows) then
+      val = self % weightWindowsMap % at(p)
+      minWgt = val(1)
+      maxWgt = val(2)
+      avWgt  = val(3)
```
Mikolaj-A-Kowalski, line 485: What will happen if the user does something stupid with a map and p will be in an invalid bin and all this entries will become 0? 

Is this what the `(maxWgt /= ZERO)` check is for below? 
If so I think it would be helpful to make a comment about this edge case.

Mikolaj-A-Kowalski (commented):
Thanks adding the weight windows. Sorry for ridiculously long review time.

CollisionOperator/CollisionProcessors/neutronCEimp_class.f90
Mikolaj-A-Kowalski: `universal variables` seems like a best place (for now)

Geometry/Fields/VectorFields/weightWindowsField_class.f90
valeriaRaffuzzi: It's ok, if the weight limits are zero simply nothing happens. Zero weights can happen if the map doesn't cover the whole geometry, or if some values in the weight windows are zero (it could happen if you produce your weight windows in an iterative way in MC, during the first iterations particles might not make it far away).

Mikolaj-A-Kowalski: Sorry! I made this comment as a note for myself when assembling a review, but I forgot to remove it!

valeriaRaffuzzi: No problem, it is a fair point!

PR #5, opened 2022-08-28
Parallel tallies

ChasingNeutrons:
Merging the parallel tallies feature.

Still need to add the option for not duplicating tallies, as @Mikolaj-A-Kowalski suggested. But that's for another PR!


Mikolaj-A-Kowalski (approved):
Looks good to me.

PR #8, opened 2022-09-02
MG majorant precalculation for DT

valeriaRaffuzzi:
Precalculating the majorant is extremely simple for MG and it buys a nice 7-8% acceleration in, for example, a 2D C5G7 problem.

The majorant here is calculated in init and saved in mgBaseNeutronDatabase, instead of adding xss at every collision.

### DIFF
--- a/NuclearData/mgNeutronData/baseMgNeutron/baseMgNeutronDatabase_class.f90
+++ b/NuclearData/mgNeutronData/baseMgNeutron/baseMgNeutronDatabase_class.f90
@@ -32,18 +32,27 @@ module baseMgNeutronDatabase_class
   !!
   !! Basic type of MG nuclear Data for neutrons
   !!
-  !! All materials in aproblem are baseMgMaterials. See its documentation for
+  !! All materials in a problem are baseMgMaterials. See its documentation for
   !! details on how the physics is handled
+  !!
+  !! Sample input dictionary:
+  !!   nucData {
+  !!     type baseMgNeutronDatabase;
+  !!     PN P0;                        // or P1
+  !!   }
   !!
   !! Public Members:
   !!   mats       -> array containing all defined materials (by matIdx)
+  !!   majorant   -> majorant xs for delta tracking
   !!   activeMats -> list of matIdxs of materials active in the problem
+  !!   nG         -> number of energy groups
   !!
   !! Interface:
   !!   nuclearDatabase interface
   !!
   type, public, extends(mgNeutronDatabase) :: baseMgNeutronDatabase
     type(baseMgNeutronMaterial), dimension(:), pointer :: mats => null()
+    real(defReal), dimension(:), allocatable           :: majorant
     integer(shortInt), dimension(:), allocatable       :: activeMats
     integer(shortInt)                                  :: nG = 0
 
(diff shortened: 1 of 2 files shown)


ChasingNeutrons (commented):
Looks pretty good to me. I'm trying (and struggling) to think if there's really any circumstance that we wouldn't want this as you could in principle make it optional - probably not worth the extra code. Now we just need to do the same for CE (which I'm sure may be a bit more painful...).

NuclearData/mgNeutronData/baseMgNeutron/baseMgNeutronDatabase_class.f90
```diff
@@ -290,12 +298,22 @@ subroutine init(self, dict, ptr, silent)
...
     end do
 
+    ! Precalculate majorant xs for delta tracking
+    allocate (self % majorant(self % nG))
+    do g = 1,self % nG
+      xs = ZERO
+      do i = 1,nMat
+        xs = max(xs, self % mats(i) % data(TOTAL_XS, g))
+      end do
+      self % majorant(g) = xs
```
Mikolaj-A-Kowalski, line 314:
```suggestion
      self % majorant(g) = maxval(self % mats(:) % data(TOTAL_XS, g)
```
But please check that it really does work. My Fortran is a bit rusty.

NuclearData/mgNeutronData/baseMgNeutron/baseMgNeutronDatabase_class.f90
```diff
@@ -124,13 +127,16 @@ function getMajorantXS(self, p) result(xs)
...
     class(particle), intent(in)                 :: p
     real(defReal)                               :: xs
-    integer(shortInt)                           :: i, idx
+    character(100), parameter :: Here = ' getMajorantXS (baseMgNeutronDatabase_class.f90)'
 
-    xs = ZERO
-    do i=1,size(self % activeMats)
-      idx = self % activeMats(i)
-      xs = max(xs, self % getTotalMatXS(p, idx))
-    end do
```
Mikolaj-A-Kowalski, line 133: I think it may be a considerable problem that in the initialisation active materials are not considered. 
Unless I am missing something (quite possible -> Friday evening ;-) ) the pre-calculated majorant is based on all defined materials, not only the materials which are actually included in the geometry.    

I would suggest moving the majorant calculation to the `activate` subroutine.

Mikolaj-A-Kowalski (changes requested):

Mikolaj-A-Kowalski:
Thanks for the development! It is really nice optimisation! 

I absolutely love the documentation and corrections to the typos :seal:

NuclearData/mgNeutronData/baseMgNeutron/baseMgNeutronDatabase_class.f90
valeriaRaffuzzi: Very good point.  I will do that!

NuclearData/mgNeutronData/baseMgNeutron/baseMgNeutronDatabase_class.f90
valeriaRaffuzzi: True Fortran magic

valeriaRaffuzzi:
Done! 
@ChasingNeutrons I don't think there's any obvious disadvantage, it doesn't take much memory or time to store it even if ST is being used. And it's so simple to do that it was a shame not having it by default!

Mikolaj-A-Kowalski (approved):

PR #11, opened 2022-10-20
Common buffer for load balancing

ChasingNeutrons:
With help from @valeriaRaffuzzi , I made a common buffer (i.e., non-thread private) for fixed source problems where there's significant multiplicity. For example, near-critical problems and problems with variance reduction + weight windows.
There's also a few typo fixes about!

### DIFF
--- a/PhysicsPackages/fixedSourcePhysicsPackage_class.f90
+++ b/PhysicsPackages/fixedSourcePhysicsPackage_class.f90
@@ -208,6 +208,10 @@ contains
           end if
 
           ! Clear out common queue
+          ! Note the apparently redundant critical sections (one here in PP, one in the dungeon).
+          ! This is to prevent the situation where two threads both enter the conditional and compete
+          ! for the final particle in the dungeon. The first thread would pop the particle while the
+          ! second would try to pop from an empty dungeon.
           if (associated(self % commonBuffer)) then
             !$omp critical
             if (.not. self % commonBuffer % isEmpty()) then
--- a/PhysicsPackages/fixedSourcePhysicsPackage_class.f90
+++ b/PhysicsPackages/fixedSourcePhysicsPackage_class.f90
@@ -79,15 +79,17 @@ module fixedSourcePhysicsPackage_class
     integer(shortInt)  :: printSource = 0
     integer(shortInt)  :: particleType
     integer(shortInt)  :: bufferSize
+    integer(shortInt)  :: bufferShift
 
     ! Calculation components
     type(particleDungeon), pointer :: thisCycle       => null()
+    type(particleDungeon), pointer :: commonBuffer    => null()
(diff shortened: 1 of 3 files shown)


valeriaRaffuzzi (approved):
This works for me! 
It's totally needed for calculations with weight windows, makes things much faster and parallel performance much better.

PhysicsPackages/fixedSourcePhysicsPackage_class.f90
```diff
@@ -189,11 +192,31 @@ subroutine cycles(self, tally, N_cycles)
...
+            cycle
+          end if
+
+          ! Clear out common queue
+          if (associated(self % commonBuffer)) then
+            !$omp critical
+            if (.not. self % commonBuffer % isEmpty()) then
+              call self % commonBuffer % releaseCritical(p)
+            end if
+            !$omp end critical
```
Mikolaj-A-Kowalski, line 216:
```suggestion
            if (.not. self % commonBuffer % isEmpty()) then
              call self % commonBuffer % releaseCritical(p)
            end if
```
The critical region is controlled inside the subroutine isn't it?

PhysicsPackages/fixedSourcePhysicsPackage_class.f90
```diff
@@ -189,11 +192,31 @@ subroutine cycles(self, tally, N_cycles)
             if(p % isDead) exit history
           end do history
 
+          ! If buffer is quite full, shift some particles to the commonBuffer
```
Mikolaj-A-Kowalski, line 195: 'quite full'  :gb: :tea: :crown:

Mikolaj-A-Kowalski (approved):
Thanks for the change! Just the one minor suggestion and I think it is ready to merge.

PhysicsPackages/fixedSourcePhysicsPackage_class.f90
ChasingNeutrons: The thing I'm being careful about here is that check on whether the buffer is empty. If there are two threads and one particle in the buffer, both will enter the conditional, the first will critically remove the particle, while the second will then be left trying to pop a particle from an empty buffer.
We could put a conditional inside the release as a check in principle? But we would still have the critical in the physics package too. What do you think?

Mikolaj-A-Kowalski: Good point! It completely eluded me. I guess the only thing that is worth changing here is maybe adding a comment to explain why it is like that. I am sure I am not a last person who might be confused by it :-)!

ChasingNeutrons: Hear hear... We only found out because we were bitten by it ourselves!

PR #13, opened 2022-12-03
Parallel efficiency in tallies

### LINKED ISSUE #12: Out-of-bounds array access in invertInelastic
I was running in debug mode (so array bounds checking switched on) and got a crash on line 198 of aceNeutronNuclide_class.f90.

Namely, inside the invertInelastic when calling 
`bottomXS = self % MTdata(i) % xs(idxT)`

This presumably happens because one or two lines previous there's a check
`if( idxT < 0 ) cycle`
which obviously misses the possibility of 0.

Nothing seems to go wrong when debug mode is disabled, but this should certainly be sorted out!

ChasingNeutrons:
A few tweaks to tallies to efficiently process them in parallel and avoid false sharing of tally cache lines with some (parameterised) array padding. Closes #12


Tallies/scoreMemory_class.f90
```diff
@@ -178,10 +181,10 @@ subroutine score_defReal(self, score, idx)
     character(100),parameter :: Here = 'score_defReal (scoreMemory_class.f90)'
 
     ! Verify bounds for the index
-    if( idx < 0_longInt .or. idx > self % N) then
-      call fatalError(Here,'Index '//numToChar(idx)//' is outside bounds of &
-                            & memory with size '//numToChar(self % N))
-    end if
+    !if( idx < 0_longInt .or. idx > self % N) then
```
valeriaRaffuzzi, line 184: Should we keep the check here?

valeriaRaffuzzi (approved):
Looks good to me! 
Thanks for fixing the bad scaling

PR #14, opened 2022-12-25
Random ray method in SCONE

ChasingNeutrons:
Here is both an eigenvalue and fixed source random ray solver (that are running pretty well). I don't know whether we have strong feelings about whether or not they should be merged in ultimately? If so, feel free to take a look and comment whenever.


ChasingNeutrons:
Closed given the neater version awaiting review.

Issue #10, opened 2022-10-03
Upgrade to pfUnit >=4.0


Mikolaj-A-Kowalski:
I have run into a problem where our gold old pfUnit python preprocessor fails with Python 3.10 [But should fail for any version >=3.8] 
The reason seems to be the fact that it cannot find the `iterable` in `collections` module, which is due to the fact that it has been deprecated. As the result we have no choice but to update to pfUnit 4.0. Due to the requirements of the latter it will mean the definite end of support for the older gfortran versions. Hopefully we will be able to preserve 8. At worst we will need to switch to 9. 

I will start looking into that and try to keep this thread updated.

ChasingNeutrons:
Resolved by #70

PR #15, opened 2023-01-10
Enable the stricter runtime sanitization checks with `-DDEBUG`

DaveDuck321:
The `hashFunctions` unit-tests failed when the extra runtime sanitation was enabled. This PR fixes the issue and enables the stricter checks in debug mode.

@ChasingNeutrons does this look reasonable?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,9 @@ if(COVERAGE)
 endif()
 
 # Add Debugging flags to scone library
+# NOTE: GFortran 7.5 may give buggy warnings here
+#   If this is an issue, consider using `-fcheck=bounds,do,mem,pointer`
+#   https://github.com/CambridgeNuclear/SCONE/pull/15#issuecomment-1384192425
 if(DEBUG)
   list(APPEND scone_extra_flags "-fcheck=all,no-array-temps" "-Waliasing")
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,7 +38,7 @@ endif()
 
 # Add Debugging flags to scone library
 if(DEBUG)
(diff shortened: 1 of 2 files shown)


Mikolaj-A-Kowalski:
Thanks for the PR! I think it is ready to go! 
The only question I have is which versions of the gfortran were you able to run it with? I just vaguely remember that I did not enable all the checks back in the day, because they used to be a big buggy and there were some false positives.

Mikolaj-A-Kowalski (approved):

DaveDuck321:
I'm using `GNU Fortran (GCC) 12.2.0`, there haven't been any false positives in the unit-tests or input files I've tried. If everyone else is using an older version maybe that should be tested too?

There is now an extra runtime warning about temporary array creation in the `uniFills` test but I've kept this on since this should probably be avoided in new code.

Mikolaj-A-Kowalski:
I have checked with the older GFortran compilers (7,8,9) and it seems that the 7.5 results in the buggy warnings. 

If you are all right with it I would just add a comment in CMake where the debug flags are set to highlight that it may not work correctly with GFortran 7.5.  Anyway, I think that GFortran 7.5 is lost to us as soon we will have to switch to pfUnit in version 4.0 (see #10 ) so it is not a problem. 

I can do it myself if you prefer. Otherwise I think it is ready to merge unless there is anything you wish to add? 

Once again thank you for the fix!

DaveDuck321:
> I can do it myself if you prefer. Otherwise I think it is ready to merge unless there is anything you wish to add?

Cool, thanks. I've added the comment you suggested.

This PR should be good to merge as a standalone change, I'm not planning to add anything else regarding the tooling.

PR #16, opened 2023-01-16
Fix format of a real in dictionary parsing

### LINKED ISSUE #6: Small fix needed in input file reading
When in the input file one writes, for example, for an energy boundary, 1E-11 rather than 1.0E-11, SCONE reads 1E-111. We should either fix this or at least have a warning when energy values are unreasonable.

Mikolaj-A-Kowalski:
Resolves #6 
Just makes a change in the format used to read real into a dictionary.


ChasingNeutrons (approved):
All looks good to me! Good spot!

PR #18, opened 2023-01-18
Cones

valeriaRaffuzzi:
Here's a cone quadratic surface. It's an infinite cone, aligned with the x, y or z axis. 
The input definition is the same as MCNP (need to specify the tangent of the opening angle and the orientation of the cone), compatible with Serpent as well. 

I wrote a unit test and compared my results with Serpent, which agree.


Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,346 @@
...
+   end select
+
+   ! Load data
+   self % t_sq   = tangent * tangent
+   self % vertex = vertex
+   self % dir = dir
+   call self % setID(id)
+
+   ! Set surface tolerance
+   call self % setTol(TWO * SURF_TOL)
```
Mikolaj-A-Kowalski, line 188: It probably should be made consistent with the value in the doc-comment. 

Sadly there is a general problem with the surfaces and surface tolerance I missed when the current geometry was first made. 

I wanted the surface tolerance to be well defined and be approximately a shell of constant 'fuzzy thickness' around the surface, but it sadly will not work with general quadratic surfaces where the 'radius' changes. Hence one cannot get away with the approximation of `2 * r * SURF_TOL` which worked well for cylinders and spheres...   

I think it is fine to leave it as it is but perhaps there should be a comment that will say that 'the thickness of the fuzzy region around the surface will depend on the position on the surface and is not constant'. 

We probably should open an issue about it as well. We need to rethink how the surface tolerance is done.

Mikolaj-A-Kowalski (commented):
Thanks for the contribution and sorry that surfaces are bit of the mess in SCONE. 

The only major question I have is if wouldn't it be easier to just include both cones in the surface instead of trying to use only one? I guess it makes it more user friendly but it seems to require considerable complication when evaluating the surface expression or finding the distance.

Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,346 @@
...
+  !!
+  !! F(r) = (ri - oi)^2 + (rj - oj)^2 - t^2 * (rk - ok)^2 = 0
+  !!
+  !! Where i,j,k are x,y & z axis, and t is the tangent of the cone opening angle
+  !! The cone is aligned along the k axis
+  !!
+  !! The cone can be oriented towards the +k or -k directions; in the input the entry
+  !! orientation should respectively be 1 or -1.
+  !!
+  !! Three diffrent types are avaliable
```
ChasingNeutrons, line 24:
```suggestion
  !! Three different types are available
```

Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,346 @@
...
+    integer(shortInt), intent(in)           :: id
+    character(*), intent(in)                :: type
+    real(defReal), dimension(3), intent(in) :: vertex
+    real(defReal), intent(in)               :: tangent
+    integer(shortInt), intent(in)           :: dir
+    character(100), parameter :: Here = 'build (cylinder_class.f90)'
+
+    ! Check values
+   if (id < 1) then
+     call fatalError(Here,'Invalid surface id provided. ID must be > 1')
```
ChasingNeutrons, line 152:
```suggestion
     call fatalError(Here,'Invalid surface id provided. ID must be > 0')
```

Geometry/Surfaces/Tests/cone_test.f90
```diff
@@ -0,0 +1,349 @@
...
+
+  !!
+  !! Test boundary conditions
+  !!
+@Test(cases=[1,2,3])
+  subroutine testBC(this)
+    class(test_cone), intent(inout) :: this
+    real(defReal), dimension(3)     :: r, u, r_pre, u_pre
+
+    ! Set Boundary Contidions
```
ChasingNeutrons, line 171:
```suggestion
    ! Set Boundary Conditions
```

Geometry/geometryStd_class.f90
```diff
@@ -414,18 +414,18 @@ function activeMats(self) result(matList)
   end function activeMats
 
   !!
-  !! Descend down the geometry structure untill material is reached
+  !! Descend down the geometry structure until material is reached
   !!
-  !! Requires strting level to be specified.
+  !! Requires starting level to be specified.
   !! It is private procedure common to all movment types in geometry.
```
ChasingNeutrons, line 420:
```suggestion
  !! It is a private procedure common to all movement types in geometry.
```

Geometry/Surfaces/Tests/cone_test.f90
```diff
@@ -0,0 +1,349 @@
...
+    u(a) = ZERO
+    u(p1) = -ONE
+    u(p2) = ZERO
+    u = u/norm2(u)
+    ref = TWO
+    @assertEqual(ref, this % surf % distance(r, u), TOL * ref)
+
+    ! Particle going outside
+    @assertEqual(INF, this % surf % distance(r, -u))
+
```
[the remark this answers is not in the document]
valeriaRaffuzzi: This should be gone now, thanks to few more checks! 

I also changed the orientation thing, and now one can truncate the cone however he/she wants. Both at the vertex, or at any other height at either side of the vertex.

Geometry/Surfaces/QuadSurfaces/cone_class.f90
valeriaRaffuzzi: The tests were a bit challenges because of this weird tolerance constant through a changing radius. Now everything should work but I agree that it's a bit annoying and suboptimal.

Hopefully this could be discussed in an issue and addressed in another PR!

Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,417 @@
...
+    else
+      d1 = - (b + sqrt(delta)) / a
+      d2 = - (b - sqrt(delta)) / a
+
+    end if
+
+    ! If one the distance found is smaller than the surface tolerance, then the particles
+    ! is on a surface: set that distance to infinity
+    if (d1 < self % surfTol()) d1 = INF
+    if (d2 < self % surfTol()) d2 = INF
```
Mikolaj-A-Kowalski, line 306: I am not sure if this approach will work I'm afraid... :-( 
The problem are particles hitting the surface at very shallow angle. They may be well within the tolerance but the distance to the surface may be still quite large. This is why in the e.g. 'cylinder' we need an extra case in the if-else for when a particle is within surface tolerance. 

Sorry... geometry is super annoying.

Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,417 @@
...
+    ! Verify if the points found would fall inside the cone
+    testPoint2 = r(self % plane) + d1 * u(self % plane) - self % vertex(self % plane)
+    testRadius = dot_product(testPoint2, testPoint2)
+    maxRadius  = self % hMax * self % hMax * self % tan2
+    if (testRadius > maxRadius) d1 = INF
+
+    testPoint2 = r(self % plane) + d2 * u(self % plane) - self % vertex(self % plane)
+    testRadius = dot_product(testPoint2, testPoint2)
+    maxRadius  = self % hMin * self % hMin * self % tan2
+    if (testRadius > maxRadius) d2 = INF
```
Mikolaj-A-Kowalski, line 334: As an optimisation we can switch to the approach similar to the one  used in the `box`... basically it involves calculating the ranges on the line that fall within a particular body (e.g. inside of the cylinder or space between a pair of planes) and then performing the intersections on the ranges. But is is sadly a bit difficult to explain in the short comment ;-( We can talk about it in person perhaps. 

That being said we can leave the explicit approach if it works well and include the alternative as a TODO in the comment.

Mikolaj-A-Kowalski (commented):
Looks good and thanks for the changes.
But because it is geometry it makes me super paranoid. Did you run a lot of surface tracking calculations with the cylinders yet? We probably should construct geometries and torture them with the `rayVol` physics package a bit before merging. 
I will try to run some when I have some time.

Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,417 @@
...
+    if (id < 1) call fatalError(Here,'Invalid surface id provided. ID must be > 0')
+
+    if (size(vertex) /= 3) then
+      call fatalError(Here,'Vertex needs to have size 3. It has size: '//numToChar(size(vertex)))
+    end if
+
+    if (angle <= ZERO .or. angle >= 90.0_defReal) then
+      call fatalError(Here, 'Opening angle of cone must be in the range 0-90 degrees. &
+                            & It is: '//numToChar(angle))
+    end if
```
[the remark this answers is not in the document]
valeriaRaffuzzi: Maybe the error message should be reworded, but the cases of `0` and `90` should be caught as an error.

Geometry/Surfaces/QuadSurfaces/cone_class.f90
valeriaRaffuzzi: I see, I hadn't thought about that! I will reintroduce the extra if-else case when solving the quadratic, hopefully that will do it!

Geometry/Surfaces/QuadSurfaces/cone_class.f90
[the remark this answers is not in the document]
Mikolaj-A-Kowalski: ... or I should learn to read better :sweat_smile: 
Sorry, I misread the if statement and though these cases are accepted.

Geometry/Surfaces/QuadSurfaces/cone_class.f90
valeriaRaffuzzi: No that's ok, over the weekend I will have a look at the box and fix this! But I would still need to calculate explicitly whether the particle would intersect the circles? Because in a box or cylinder the two faces are of the same size but this is not true here.

Geometry/Surfaces/QuadSurfaces/cone_class.f90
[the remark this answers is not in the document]
valeriaRaffuzzi: No worries, I can add an explicit comment saying that the degenerate cases of 0 and 90 are not acceptable.

valeriaRaffuzzi:
> Looks good and thanks for the changes. But because it is geometry it makes me super paranoid 😨 Did you run a lot of surface tracking calculations with the cones yet? We probably should construct geometries and torture them with the `rayVol` physics package a bit before merging. I will try to run some when I have some time.

Thanks a lot for reviewing this, I am a geometry beginner! I ran a couple of cases but not `rayVol` to be fair. I would say hold on with this, I will work on it a bit more during the weekend and fix everything you mentioned, and then do a bit more testing :)

Geometry/Surfaces/QuadSurfaces/cone_class.f90
valeriaRaffuzzi: Ok, the new distance calculation should be more like box and truncated cylinder now. The only thing I had to add is a check to make sure that the intersection found is on the right half-cone. Otherwise, the segment between two intersections could have been outside of the cone. If an intersection is on the wrong side of the cone, the distance is set to + or - infinity in a way that the intersection 'segment' (or ray) crosses the truncated cone (when here is an actual crossing). This might not be clear, in case let me know!

valeriaRaffuzzi:
> > Looks good and thanks for the changes. But because it is geometry it makes me super paranoid 😨 Did you run a lot of surface tracking calculations with the cones yet? We probably should construct geometries and torture them with the `rayVol` physics package a bit before merging. I will try to run some when I have some time.
> 
> Thanks a lot for reviewing this, I am a geometry beginner! I ran a couple of cases but not `rayVol` to be fair. I would say hold on with this, I will work on it a bit more during the weekend and fix everything you mentioned, and then do a bit more testing :)
> 
> Hopefully at least we're converging to something with these cones!

I think `rayVol` might be broken. I tried to use it on regular JEZ and POPSY and I get an error: `Coordinate list is not placed in the geometry`, in `move`. We should open an issue about that. However, I tested the cones on a few configurations doing ST vs DT and they always agree perfectly! I also tried to make the unit test as varied as I could in terms of distance calculations. Let me know if there is anything else I could do to test this.

Geometry/Surfaces/QuadSurfaces/cone_class.f90
valeriaRaffuzzi: I changed the tolerance to be proportional to the 'mean radius' of the cone now. That is the radius of the cone at mid-height.

Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,451 @@
...
+    ! Check values
+    if (id < 1) call fatalError(Here,'Invalid surface id provided. ID must be > 0')
+
+    if (size(vertex) /= 3) then
+      call fatalError(Here,'Vertex needs to have size 3. It has size: '//numToChar(size(vertex)))
+    end if
+
+    if (angle <= ZERO .or. angle >= 90.0_defReal) then
+      call fatalError(Here, 'Opening angle of cone must be in the range 0-90 degrees &
+                            & (extremes excluded). It is: '//numToChar(angle))
```
ChasingNeutrons, line 144: Maybe my Fortran is getting rusty, shouldn't this be:
```suggestion
                    call fatalError(Here, 'Opening angle of cone must be in the range 0-90 degrees '//&
                            '(extremes excluded). It is: '//numToChar(angle))
```

Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,451 @@
...
+
+    ! Check values
+    if (id < 1) call fatalError(Here,'Invalid surface id provided. ID must be > 0')
+
+    if (size(vertex) /= 3) then
+      call fatalError(Here,'Vertex needs to have size 3. It has size: '//numToChar(size(vertex)))
+    end if
+
+    if (angle <= ZERO .or. angle >= 90.0_defReal) then
+      call fatalError(Here, 'Opening angle of cone must be in the range 0-90 degrees &
```
ChasingNeutrons, line 143: Maybe my Fortran is getting rusty, shouldn't this be:
```suggestion
                    call fatalError(Here, 'Opening angle of cone must be in the range 0-90 degrees '//&
                            '(extremes excluded). It is: '//numToChar(angle))
```

Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,451 @@
...
+                            & (extremes excluded). It is: '//numToChar(angle))
+    end if
+
+    if (hMin >= hMax) call fatalError(Here, 'hMin is greater or equal hMax.')
```
ChasingNeutrons, line 147:
```suggestion
    if (hMin >= hMax) call fatalError(Here, 'hMin is greater than or equal to hMax.')
```

Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,451 @@
...
+  pure function boundingBox(self) result(aabb)
+    class(cone), intent(in)     :: self
+    real(defReal), dimension(6) :: aabb
+    real(defReal)               :: maxRadius
+
+    ! Along the axis
+    aabb(self % axis) = self % vertex(self % axis) + self % hMin
+    aabb(self % axis + 3) = self % vertex(self % axis) + self % hMax
+
+    ! On the plane of the basis
```
ChasingNeutrons, line 224:
```suggestion
    ! On the plane of the base
```

Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,451 @@
...
+
+  !!
+  !! Evaluate surface expression c = F(r)
+  !!
+  !! See surface_inter for details
+  !!
+  pure function evaluate(self, r) result(c)
+    class(cone), intent(in)                 :: self
+    real(defReal), dimension(3), intent(in) :: r
+    real(defReal)                           :: c, cMin, cMax, cBasis
```
ChasingNeutrons, line 239:
```suggestion
    real(defReal)                           :: c, cMin, cMax, cBase
```

Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,451 @@
...
+    diff = r - self % vertex
+
+    ! Evaluate the expression for the surface of the cone
+    c = dot_product(diff(self % plane), diff(self % plane)) - &
+        self % tanSquare * diff(self % axis) ** 2
+
+    ! Evaluate the expressions for the two basis of the cone and find the maximum
+    cMin = - diff(self % axis) + self % hMin
+    cMax = diff(self % axis) - self % hMax
+    cBasis = max(cMin, cMax)
```
ChasingNeutrons, line 252:
```suggestion
    cBase = max(cMin, cMax)
```

Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,451 @@
...
+
+    ! Find the overall maximum
+    c = max(c, cBasis)
```
ChasingNeutrons, line 255:
```suggestion
    c = max(c, cBase)
```

Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,451 @@
...
+  pure function evaluate(self, r) result(c)
+    class(cone), intent(in)                 :: self
+    real(defReal), dimension(3), intent(in) :: r
+    real(defReal)                           :: c, cMin, cMax, cBasis
+    real(defReal), dimension(3)             :: diff
+
+    ! Vector for the difference between position provided and vertex
+    diff = r - self % vertex
+
+    ! Evaluate the expression for the surface of the cone
```
ChasingNeutrons, line 245: All of the cone is arguably a surface.
```suggestion
    ! Evaluate the expression for the slanted surface of the cone
```

Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,451 @@
...
+
+    ! Calculate quadratic components in the cone
+    a = dot_product(u(self % plane), u(self % plane)) - &
+        self % tanSquare * u(self % axis)**2
+    b = dot_product(diff(self % plane) , u(self % plane)) - &
+        self % tanSquare * diff(self % axis) * u(self % axis)
+    c = dot_product(diff(self % plane), diff(self % plane)) - &
+        self % tanSquare * diff(self % axis) ** 2
+
+    ! Calculate delta/4
```
ChasingNeutrons, line 296: What is delta? You mean the discriminant? That also does not appear to be what is calculated here - that would be b^2 / 4 - a*c

Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,451 @@
...
+    delta = b * b - a * c
+
+    ! Calculate the distances from the cone surface
```
ChasingNeutrons, line 299:
```suggestion
    ! Calculate the distances from the slanted surface
```

Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,451 @@
...
+    ! Calculate the distances from the basis of the cone
+    if (abs(u(self % axis)) > epsilon(ONE)) then    ! Normal intersection
+
+      d1 = (self % hMax - diff(self % axis)) / u(self % axis)
+      d2 = (self % hMin - diff(self % axis)) / u(self % axis)
+
+    else                                            ! Particle parallel to axis: check location
+
+      if (diff(self % axis) > self % hMin .and. diff(self % axis) < self % hMax) then
+        ! Inside the cone: basis intersection segment is -INF:+INF
```
ChasingNeutrons, line 340: Should this be distance rather than segment?
```suggestion
        ! Inside the cone: base intersection distance is -INF:+INF
```

Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,451 @@
...
+        d1 = -INF
+        d2 = INF
+      else
+        ! Outside the cone: basis intersection segment doesn't exist
```
ChasingNeutrons, line 344:
```suggestion
        ! Outside the cone: base intersection distance doesn't exist
```

Geometry/Surfaces/QuadSurfaces/cone_class.f90
```diff
@@ -0,0 +1,451 @@
...
+    class(cone), intent(in)                 :: self
+    real(defReal), dimension(3), intent(in) :: r
+    real(defReal), dimension(3), intent(in) :: u
+    logical(defBool)                        :: halfspace
+    real(defReal), dimension(3)             :: diff, norm
+    real(defReal)                           :: proj
+
+    diff = r - self % vertex
+
+    ! Check the location of the particle, i.e., basis or cone surface, to calculate
```
ChasingNeutrons, line 402:
```suggestion
    ! Check the location of the particle, i.e., base or cone surface, to calculate
```

Geometry/Surfaces/Tests/cone_test.f90
```diff
@@ -0,0 +1,481 @@
...
+    u     = ZERO
+    u(a)  = ONE
+    u(p1) = -ONE
+    u = u / norm2(u)
+
+    ! At the surface
+    @assertFalse(this % surf % halfspace(r, u))
+
+    ! Out within tolerance
+    ! NOTE THAT the tolerance is very sensitive: if using eps = HALF * TOL, the
```
ChasingNeutrons, line 235:
```suggestion
    ! NOTE that the tolerance is very sensitive: if using eps = HALF * TOL, the
```

Geometry/Surfaces/Tests/cone_test.f90
```diff
@@ -0,0 +1,481 @@
...
+
+    ! Out outside tolerance
+    eps = -TWO * tolerance
+    @assertTrue(this % surf % halfspace(r + eps*u, u))
+
+    ! Well Outside
+    eps = -HALF
+    @assertTrue(this % surf % halfspace(r + eps*u, u))
+
+    ! Well withn
```
ChasingNeutrons, line 248:
```suggestion
    ! Well within
```

Geometry/Surfaces/Tests/cone_test.f90
```diff
@@ -0,0 +1,481 @@
...
+    eps = HALF
+    @assertFalse(this % surf % halfspace(r + eps*u, u))
+
+    ! Tangent particle should be outside
+    u     = ZERO
+    u(p2) = -ONE
+    u = u /norm2(u)
+    @assertTrue(this % surf % halfspace(r, u))
+
+    ! Choose point on one of the basis of the cone, moving out
```
ChasingNeutrons, line 258:
```suggestion
    ! Choose point on one of the bases of the cone, moving out
```

Geometry/Surfaces/Tests/cone_test.f90
```diff
@@ -0,0 +1,481 @@
...
+
+    ! Get surface tolerance
+    tolerance = this % surf % surfTol()
+
+    ! Set axis and plane axis indices
+    a = this % axis
+    p1 = this % plane(1)
+    p2 = this % plane(2)
+
+    ! **Outside the cone
```
ChasingNeutrons, line 311:
```suggestion
    ! Outside the cone
```

Geometry/Surfaces/Tests/cone_test.f90
```diff
@@ -0,0 +1,481 @@
...
+    r(p2) = ONE
+
+    ! Impact at an angle, direction of flight on an axis
+    u(a)  = ONE
+    u(p1) = ZERO
+    u(p2) = ZERO
+    ref   = 4.0_defReal
+    @assertEqual(ref, this % surf % distance(r, u), TOL * ref)
+
+    ! Impacting the basis
```
ChasingNeutrons, line 323:
```suggestion
    ! Impacting the bases
```

Geometry/Surfaces/Tests/cone_test.f90
```diff
@@ -0,0 +1,481 @@
...
+    @assertEqual(INF, this % surf % distance(r, u))
+
+    ! Going in the opposite direction
+    u(a)  = -ONE
+    u(p1) = ZERO
+    u(p2) = ZERO
+    u   = u/norm2(u)
+    @assertEqual(INF, this % surf % distance(r, u))
+
+    ! Parallel to the basis
```
ChasingNeutrons, line 345:
```suggestion
    ! Parallel to the bases
```

Geometry/Surfaces/Tests/cone_test.f90
```diff
@@ -0,0 +1,481 @@
...
+
+    ! Parallel to axis
+    u(a)  = ONE
+    u(p1) = ZERO
+    u(p2) = ZERO
+    u = u/norm2(u)
+    ref = 3.0_defReal
+    @assertEqual(ref, this % surf % distance(r, u), TOL * ref)
+
+    ! Parallel to basis
```
ChasingNeutrons, line 404:
```suggestion
    ! Parallel to bases
```

Geometry/geometryStd_class.f90
```diff
@@ -419,18 +419,18 @@ function activeMats(self) result(matList)
   end function activeMats
 
   !!
-  !! Descend down the geometry structure untill material is reached
+  !! Descend down the geometry structure until material is reached
   !!
-  !! Requires strting level to be specified.
-  !! It is private procedure common to all movment types in geometry.
+  !! Requires starting level to be specified.
+  !! It is private procedure common to all movement types in geometry.
```
ChasingNeutrons, line 425:
```suggestion
  !! It is a private procedure common to all movement types in geometry.
```

docs/User Manual.rst
```diff
@@ -517,6 +517,23 @@ Example: ::
 
       billy { id 92; type xCylinder; origin (0.0 0.0 9.0); radius 4.8; }
 
+* cone: cone aligned with x, y or z axis, and truncated arbitrarily on both sides. 
+  The input type has to be ``xCone``, ``yCone`` or ``zCone``. The orientation of the
```
ChasingNeutrons, line 521: Orientation is ambiguous with x, y, and z
```suggestion
  The input type has to be ``xCone``, ``yCone`` or ``zCone``. The gradient of the
```

docs/User Manual.rst
```diff
@@ -517,6 +517,23 @@ Example: ::
...
+    surface. [degrees]
+  - hMin: coordinate along the cone axis corresponding to the position of the
+    cone lower basis with respect to the vertex. It could be positive or negative. [cm]
+  - hMax: coordinate along the cone axis corresponding to the position of the
+    cone upper basis with respect to the vertex. It could be positive or negative.
+    Must be of the same sign as hMin, and always larger. [cm]
+
+Example: ::
+
+      billy { id 92; type xCone; vertex (1.1 4.0 2.98); angle 30; hMin 5.0; hMax 15.0; }
```
ChasingNeutrons, line 535:
```suggestion
      connor { id 92; type xCone; vertex (1.1 4.0 2.98); angle 30; hMin 5.0; hMax 15.0; }
```

ChasingNeutrons (approved):
All small things I think. Should be straightforward.

docs/User Manual.rst
valeriaRaffuzzi: Did you change name to Connor because it sounds like cone?

ChasingNeutrons: Yes

valeriaRaffuzzi: Fair enough. I love bad humor

Geometry/Surfaces/QuadSurfaces/cone_class.f90
ChasingNeutrons: Sure, but it isn't clear what delta is. If you say 'calculate delta/4' there is no way for someone reading afterwards to understand what you are referring to.

Geometry/Surfaces/QuadSurfaces/cone_class.f90
valeriaRaffuzzi: I think segment is ok: each couple of distances (base1 and base2, or slanted1 and slanted2) identifies 2 points that create a segment. The intersection between the two segments is what is actually internal to the cone. This is what the code is doing, and I think the docs are consistent with the cylinder and box where this approach is used.

ChasingNeutrons: I would then say the segment lies between -INF:INF. It is not clear what a segment being -INF:INF means.

Geometry/Surfaces/QuadSurfaces/cone_class.f90
valeriaRaffuzzi: Should I call it discriminant instead? Is this what you mean?

ChasingNeutrons: Yes, that would be easier to follow.

PR #21, opened 2023-01-19
Update SCONE ACE Library file

Mikolaj-A-Kowalski:
This PR makes couple of modifications to the ACE library file. 
- Delimiters are changed from space to semi-colon to allow paths with spaces 
- Limitation on the length of absolute path to an ACE file is removed (in practice determined only by the size of the buffer character to read lines [900 characters at the moment?]) 
- Adds a bash script to search ASCII files for first lines of ACE cards, which writes the library file in SCONE's format. 

If this PR is integrated it will of course break backwards compatibility with existing 'Library Files'. However thanks to the script generating new ones should not be the problem. 

One limitation is that path to the library file itself still cannot contain a space (as a space would confuse the input dictionary parsing... but technically it should be circumvented by using the environmental variable to specify library file path)

### DIFF
--- a/NuclearData/ceNeutronData/aceLibrary_mod.f90
+++ b/NuclearData/ceNeutronData/aceLibrary_mod.f90
@@ -40,10 +40,10 @@ module aceLibrary_mod
   !! Storage for data associated with each library entry
   !!
   type, private :: item
-    character(nameLen) :: ZAID
-    integer(shortInt)  :: firstLine = 1
-    integer(shortInt)  :: type = UNDEF
-    character(pathLen) :: path
+    character(nameLen)        :: ZAID
+    integer(shortInt)         :: firstLine = 1
+    integer(shortInt)         :: type = UNDEF
+    character(:), allocatable :: path
   end type
 
 !!<><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
@@ -52,7 +52,7 @@ module aceLibrary_mod
 
   type(item),dimension(:),allocatable :: entry
   type(charMap)                       :: map
-  character(pathLen)                  :: libFile
+  character(:), allocatable           :: libFile
 
   public :: load
   public :: new_neutronACE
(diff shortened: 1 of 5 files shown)


NuclearData/ceNeutronData/aceLibrary_mod.f90
```diff
@@ -1,8 +1,15 @@
 !!
 !! This module is a factory that produces aceCards for a given nuclide
 !!
+!! Two types of ACE cards can be specified
+!!   Continious energy neutron data e.g. ZZAAA.TTc
+!!   SAB bard e.g. XXXXX.TT.t
```
valeriaRaffuzzi, line 6: card*

docs/Installation.rst
```diff
@@ -261,14 +261,27 @@ its own library file. An example of it is given in *IntegrationTestFiles/testLib
...
-  92233.03c  1       <absolute_path>/9233JEF33.ace
-  1001.03c   4069    <absolute_path>/1001JEF33.ace
+  92233.03c;  1;       <absolute_path>/9233JEF33.ace;
+  1001.03c;   4069;    <absolute_path>/1001JEF33.ace;
   ...
 
 `Line Number` is the line in the file at which a particular data card begins. Each line cannot
-contain more then a single entry. Each component must be space separated. Path can have only 100
-character and contain no spaces.
+contain more then a single entry. Each component must be delimited by a semi-colon.
```
valeriaRaffuzzi, line 269: than*

valeriaRaffuzzi (approved):
Looks great, things are neater and the script to produce the libraries was well needed! 
I just commented on a couple of typos

NuclearData/ceNeutronData/aceLibrary_mod.f90
```diff
@@ -114,19 +122,19 @@ subroutine load(path)
 
       ! Read if line is not empty
       if( len_trim(buffor) /= 0) then
-        associate ( bounds => splitChar(buffor, ' ') )
-          if (size(bounds,2) < 3) call fatalError(Here, 'Ill formatted line: '//trim(buffor))
+        associate ( bounds => splitChar(buffor, DELIM) )
```
ChasingNeutrons, line 125: Another typo: buffer*

PR #22, opened 2023-01-22
Uniform Fission Sites

valeriaRaffuzzi:
This pull request includes a few changes. 

First, I am adding UFS as a vector field. This requires the specification of a spatial map over some fissile material. The input information are read inside the eigenPP; additionally, a UFS flag must be specified in the implicit collision operator for this to work. 

This spatial/material map does not have to be homogeneous in terms of fissile material volume; if it isn't, one should specify it in the input with a 'uniformMap 0' flag. In this case, a simple routine to estimate the fissile material volume of each bin in the map is run during initialisation. 

Then, when generating fission neutrons in collOpImp, the map is read and updated. 

Additional features in this pull request are: 
- adding unstruct and equivolume bins in the tally sphericalMap
- generating a 3D cylindrical tally map, i.e. with radial, axial and azimuthal bins

### DIFF
--- a/CollisionOperator/CollisionProcessors/neutronCEimp_class.f90
+++ b/CollisionOperator/CollisionProcessors/neutronCEimp_class.f90
@@ -346,7 +346,7 @@ contains
     type(neutronMicroXSs)                :: microXSs
     type(fissionCE), pointer             :: fiss
     type(particleState)                  :: pTemp
-    real(defReal),dimension(3)           :: r, dir
+    real(defReal),dimension(3)           :: r, dir, val
     integer(shortInt)                    :: n, i
     real(defReal)                        :: wgt, rand1, E_out, mu, phi
     real(defReal)                        :: sig_nufiss, sig_fiss, k_eff
--- a/CollisionOperator/CollisionProcessors/neutronCEimp_class.f90
+++ b/CollisionOperator/CollisionProcessors/neutronCEimp_class.f90
@@ -26,6 +26,10 @@ module neutronCEimp_class
   use neutronScatter_class,          only : neutronScatter, neutronScatter_TptrCast
   use fissionCE_class,               only : fissionCE, fissionCE_TptrCast
 
+  ! Geometry and fields
+  use geometryReg_mod,                only : gr_fieldIdx => fieldIdx, gr_fieldPtr => fieldPtr
+  use uniFissSitesField_class,        only : uniFissSitesField, uniFissSitesField_TptrCast
+
   ! Cross-Section Packages
   use neutronXsPackages_class,       only : neutronMicroXSs
 
@@ -51,6 +55,7 @@ module neutronCEimp_class
   !!  avgWgt  -> weight of a particle on surviving splitting (optional)
   !!  impAbs  -> is implicit capture performed? (off by default)
   !!  impGen  -> are fission sites generated implicitly? (on by default)
+  !!  UFS     -> uniform fission sites variance reduction
   !!  splitting -> splits particles above certain weight (on by default)
   !!  roulette  -> roulettes particles below certain weight (off by defautl)
   !!  thresh_E -> Energy threshold for explicit treatment of target nuclide movement [-].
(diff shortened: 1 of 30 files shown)


CollisionOperator/CollisionProcessors/neutronCEimp_class.f90
```diff
@@ -485,10 +517,20 @@ subroutine split(self, p, thisCycle)
...
+    n = floor(p % w/self % maxWgt)
+    prob = p % w/self % maxWgt - n
+
+    if (p % pRNG % get() < prob) then
+      mult = n + 1
+    else
+      mult = n
+    end if
+
+    if (mult == 1) return
```
Mikolaj-A-Kowalski, line 532: I have a vague recollection that similar change was in the weight windows PR and we decided to stay with the `ceiling` approach. Or am I forgetting something?

CollisionOperator/CollisionProcessors/neutronCEimp_class.f90
```diff
@@ -168,6 +179,13 @@ subroutine init(self, dict)
          'Must generate fission sites implicitly when using implicit absorption')
     end if
 
+    ! Sets up the uniform fission sites field
+    if (self % uniFissSites) then
+      name = 'UFS'
```
Mikolaj-A-Kowalski, line 184: Similarly to weight windows, perhaps we should make the common name a constant in `universalVariables`?

Mikolaj-A-Kowalski (commented):
I think it looks all right. Thanks for the development! 

It might be good idea though to rebase it on current `main` so the GitHub Actions tries to compile and test it before we marge.

CollisionOperator/CollisionProcessors/neutronCEimp_class.f90
valeriaRaffuzzi: No, I think I changed the two branches the same way but I'll go back to normal!

valeriaRaffuzzi:
I addressed all the comments. 

However, we still need to fix the issue of the circular dependencies. I will do that in the near future, and then merge this and resume the VarReduction pull request.

Geometry/fieldFactory_func.f90
```diff
@@ -0,0 +1,80 @@
+!!
+!! Module to build new fields. Called by geometryReg_mod
+!!
```
Mikolaj-A-Kowalski, line 3: I think it is now the other way around, right? 
This module calls `geomReg` and adds the geometry.

Geometry/fieldFactory_func.f90
```diff
@@ -0,0 +1,80 @@
...
+  !! Allocates and initialises field from dictionary
+  !!
+  !! Args:
+  !!   dict [in]   -> Dictionary with definition
+  !!   name [in]   -> Name of the field for the geometry registry
+  !!
+  !! Errors:
+  !!   fatalError is type of field is unknown
+  !!
+  subroutine new_field(dict, name)
```
Mikolaj-A-Kowalski, line 43: Maybe it is worth to explicitly mention that the field gets allocated inside the `geometryReg`?

Geometry/geometryFactory_func.f90
```diff
@@ -0,0 +1,85 @@
...
+  !!
+  !! Args:
+  !!   dict [in]  -> Dictionary with geometry definition
+  !!   name [in]  -> Name of the geometry for the geometry registry
+  !!   silent [in] -> Optional. Set to .true. to surpress console messeges. Default .false.
+  !!     Note that errors will still be printed with silent=.true.
+  !!
+  !! Errors:
+  !!   fatalError is type of geometry is unknown
+  !!
```
Mikolaj-A-Kowalski, line 44: Same as for fields.

Geometry/geometryFactory_func.f90
```diff
@@ -0,0 +1,85 @@
+!!
+!! Module to build new fields. Called by geometryReg_mod
+!!
```
Mikolaj-A-Kowalski, line 3: Same as for fields.

Geometry/geometryReg_mod.f90
```diff
@@ -106,23 +96,13 @@ module geometryReg_mod
...
-    class(dictionary), intent(in)          :: dict
-    logical(defBool), optional, intent(in) :: silent
-    integer(shortInt)                      :: idx
-    logical(defBool)                       :: silent_l
+  subroutine addGeom(geom, name)
+    class(geometry), allocatable, intent(inout) :: geom
+    character(nameLen), intent(in)              :: name
+    integer(shortInt)                           :: idx
     integer(shortInt), parameter :: NOT_PRESENT = -7
     character(100), parameter    :: Here = 'addGeom (geometryReg_mod.f90)'
```
Mikolaj-A-Kowalski, line 104: The doc comment needs to be updated.

Geometry/geometryReg_mod.f90
```diff
@@ -224,10 +206,10 @@ end function geomNum
...
-  subroutine addField(name, dict)
-    character(nameLen), intent(in)         :: name
-    class(dictionary), intent(in)          :: dict
-    integer(shortInt)                      :: idx
+  subroutine addField(kentta, name)
+    class(field), allocatable, intent(inout) :: kentta
+    character(nameLen), intent(in)           :: name
+    integer(shortInt)                        :: idx
     integer(shortInt), parameter :: NOT_PRESENT = -7
     character(100), parameter    :: Here = 'addField (geometryReg_mod.f90)'
```
Mikolaj-A-Kowalski, line 214: Again doc comment diverged from the subroutine.

Mikolaj-A-Kowalski (approved):
I think all it needs is to update the documentation comments and it is safe to be merged. 
Thanks for the update!

PR #23, opened 2023-02-02
MG xs tally clerk

valeriaRaffuzzi:
Tally Clerk to calculate MG cross sections. 

It requires an energy tally Map as an input. Additionally, a spatial Map can be provided (otherwise it will calculate MG xss for the whole geometry) and a flag PN: if true, also the matrices P2 -> P7 are calculated, otherwise only up to P1. 

The transport cross section is calculated too, with both the out-scatter approximation and the flux-limited.  

Unit test is also included.


Tallies/TallyClerks/MGxsClerk_class.f90
```diff
@@ -0,0 +1,868 @@
...
+
+          ! Score outgoing scattering angle for P6 matrix
+          mu = (231.0_defReal*muL**6 - 315.0_defReal*muL**4 + 105.0_defReal*muL*muL &
+                - 5.0_defReal)/16.0_defReal * preColl % wgt
+          call mem % score(mu, addr + SCATT_EV_idx + 7*self % energyN + binEnOut)
+
+          ! Score outgoing scattering angle for P7 matrix
+          mu = (429.0_defReal*muL**7 - 693.0_defReal*muL**5 + 315.0_defReal*muL**3 &
+                - 35.0_defReal*muL)/16.0_defReal * preColl % wgt
+          call mem % score(mu, addr + SCATT_EV_idx + 8*self % energyN + binEnOut)
```
ChasingNeutrons, line 372: I may have spent too much time in deterministic land where this actually matters, but it might be a bit faster to precompute the powers up to 7 (or maybe just 3 or 4). It would save quite a lot of repeated multiplication/exponentiation.

Tallies/TallyClerks/MGxsClerk_class.f90
```diff
@@ -0,0 +1,868 @@
...
+
+        ! Calculate scattering matrices and uncertainties
+        idx = N*(i-1) + j
+        if (P0 == ZERO) then
+          P0_res(1:2, idx) = ZERO
+          P1_res(1:2, idx) = ZERO
+          prod_res(1, idx) = ONE
+          prod_res(2, idx) = ZERO
+        else
+          P0_res(1, idx) = P0/scattProb*scattXS
```
ChasingNeutrons, line 548: I probably don't understand this well enough but... Shouldn't P0 just be 1? I say because it feels like neutron weight is being counted multiple times with P0 and scattXS which doesn't feel intuitive to me.

Tallies/TallyClerks/MGxsClerk_class.f90
```diff
@@ -0,0 +1,868 @@
...
+    end do
+
+  end subroutine processPN
+
+  !!
+  !! Write contents of the clerk to output file
+  !!
+  !! See tallyClerk_inter for details
+  !!
+  !! NOTE: the energy groups are ordered from thermal to fast (G_end -> G_1)
```
ChasingNeutrons, line 693: Any possibility we can flip this round? Just because index 1 being the fastest group is so ubiquitous when doing postprocessing/using other codes. It would just take traversing the do loops in the opposite direction I think

ChasingNeutrons (commented):
Looks very good to me! Just a few small comments.

Tallies/TallyClerks/CMakeLists.txt
```diff
@@ -9,15 +9,17 @@ add_sources(./tallyClerk_inter.f90
             ./keffImplicitClerk_class.f90
             ./simpleFMClerk_class.f90
             ./dancoffBellClerk_class.f90
-	    ./shannonEntropyClerk_class.f90
-	    ./centreOfMassClerk_class.f90
+	          ./shannonEntropyClerk_class.f90
+	          ./centreOfMassClerk_class.f90
```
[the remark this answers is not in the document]
valeriaRaffuzzi: I think this is an artefact of github. This doesn't look like this in my local file

Tallies/TallyClerks/MGxsClerk_class.f90
valeriaRaffuzzi: Done!

Tallies/TallyClerks/CMakeLists.txt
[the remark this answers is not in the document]
Mikolaj-A-Kowalski: Tabs are spawn of a devil... 
There is a tab at the beginning of a line. GitHub must be using different value than 4 spaces when rendering tabs.

ChasingNeutrons (approved):

ChasingNeutrons (approved):

Tallies/TallyClerks/MGxsClerk_class.f90
```diff
@@ -0,0 +1,887 @@
...
+                                   NUBAR_idx     = 5 ,&  ! NuBar
+                                   CHI_idx       = 6 ,&  ! Fission neutron spectrum
+                                   SCATT_EV_idx  = 7     ! Analog: number of scattering events
+
+  !!
+  !! Multi-group macroscopic cross section calculation
+  !!
+  !! It prints out:
+  !! capture xs, fission xs, transport xs, nu, chi, the P0 and P1 scattering matrices,
+  !! the P0 scattering production matrix. On request, also the P2 -> P7 scattering matrices.
```
Mikolaj-A-Kowalski, line 47: Is there any particular reason why the option between P1 and P7 is binary? 
I mean it should be possible for the user to specify the highest moment which is required. I don't think it needs to be a part of this PR necessary but I would be inclined to open an issue about letting N be user-defined parameter after the PR is merged.

Tallies/TallyClerks/MGxsClerk_class.f90
```diff
@@ -0,0 +1,887 @@
...
+    ! Kill and deallocate maps
+    call self % energyMap % kill()
+
+    if (allocated(self % spaceMap)) then
+      call self % spaceMap % kill()
+    end if
+
+    ! Reset memory width
+    self % width = 0
+
```
Mikolaj-A-Kowalski, line 171:
```suggestion
    self % matN = 0
    self % energyN = 0
    self % width = 0
    self % PN = .false.
```

Tallies/TallyClerks/MGxsClerk_class.f90
```diff
@@ -0,0 +1,887 @@
...
+    private
+    ! Maps
+    class(tallyMap), allocatable :: spaceMap
+    class(tallyMap), allocatable :: energyMap
+
+    ! Useful data
+    integer(shortInt) :: energyN
+    integer(shortInt) :: matN
+    integer(shortInt) :: width
+    logical(defBool)  :: PN
```
Mikolaj-A-Kowalski, line 90:
```suggestion
    integer(shortInt) :: energyN = 0 
    integer(shortInt) :: matN = 0
    integer(shortInt) :: width = 0
    logical(defBool)  :: PN = .false.
```
It is better to be explicit. For one I do not quite remember how Fortran default initialised bools ;-)

Tallies/TallyClerks/MGxsClerk_class.f90
```diff
@@ -0,0 +1,887 @@
...
+      case(N_4N)
+        score = 4.0_defReal
+      case default
+        score = ONE
+    end select
+
+    ! Score in case of scattering events
+    select case(MT)
+      case ( N_N_ELASTIC, N_N_INELASTIC, N_Nl(1):N_Nl(40), N_Ncont,         &
+             N_2N, N_2Na, N_2Nd, N_2Nf, N_2Np, N_2N2a, N_2Nl(1):N_2Nl(16),  &
```
Mikolaj-A-Kowalski, line 306: Out of my personal curiosity: 
Do `N_2N1` - `N_2N16` ever get sampled in transport calculations? 
I always though they were covered with MT=16

Tallies/TallyClerks/MGxsClerk_class.f90
```diff
@@ -0,0 +1,887 @@
...
+
+    name = 'prod'
+    call outFile % startArray(name, resArrayShape)
+    do i=1,product(resArrayShape)
+      call outFile % addResult(prod(1,i),prod(2,i))
+    end do
+    call outFile % endArray()
+
+    ! Deallocate from memory
+    deallocate(capt, fiss, transFL, transOS, nu, chi, P0, P1, prod)
```
Mikolaj-A-Kowalski, line 816:
```suggestion
```
See below.

Mikolaj-A-Kowalski (approved):
Looks all right. Thanks for the development.

Tallies/TallyClerks/MGxsClerk_class.f90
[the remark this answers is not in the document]
valeriaRaffuzzi: Yes you're right! When I created it having lower case mg looked weird to me, but I'm all for consistency

Tallies/TallyClerks/MGxsClerk_class.f90
valeriaRaffuzzi: It was mainly laziness, inspired by Serpent but with an additional high moments-low moments check

Tallies/TallyClerks/MGxsClerk_class.f90
```diff
@@ -0,0 +1,887 @@
...
+
+      name = 'P7'
+      call outFile % startArray(name, resArrayShape)
+      do i = 1,product(resArrayShape)
+        call outFile % addResult(P7(1,i),P7(2,i))
+      end do
+      call outFile % endArray()
+
+      ! Deallocate from memory
+      deallocate(P2, P3, P4, P5, P6, P7)
```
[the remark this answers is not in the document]
valeriaRaffuzzi: Does it do at the end of the calculation only? I added this line because MGxss can get very expensive, so I wanted to free some memory in between printing in the output file the xss and the scattering matrices. This is also because the output file itself takes a lot of memory (with the huge tape written all at once), and I had my simulation failing while writing the file in the past. I would keep this here for now, at least until we modify the output file writing.

Tallies/TallyClerks/MGxsClerk_class.f90
valeriaRaffuzzi: I meant this line

Tallies/TallyClerks/MGxsClerk_class.f90
Mikolaj-A-Kowalski: We will just need to open an issue about it sice I imagine that being able to reduce the number of tallied moments could help with memory consumption.

Mikolaj-A-Kowalski:
I think it is all good to merge, isn't it?

Tallies/TallyClerks/MGxsClerk_class.f90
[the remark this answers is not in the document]
valeriaRaffuzzi: I will quickly add this comment, and then I think we can go ahead and merge!

Issue #26, opened 2023-02-10
Memory reference error with JSON output


amgc500:
I'm trying to run an example with JSON output, not Matlab output, but including the line

`outputFormat asciiJSON;
`
in my input file results in the error message:

`Program received signal SIGSEGV: Segmentation fault - invalid memory reference.`

`Backtrace for this error:`
`#0  0x100baf187`
`#1  0x100bae163`
`#2  0x19ef782a3`
`#3  0x100867f4f`
`zsh: segmentation fault  ./Build/scone.out InputFiles/c5g7_mox
`

I'm running on a mac, and it runs fine with either no output format reference in the input file, or a matlab output.

Mikolaj-A-Kowalski:
Hi thanks for reporting the problem! 
I had a very quick look and sadly I was not able to reproduce the problem with my gfortran 10 on my Debian laptop :-/ 
If I remember correctly you were using gfortran 12? Sadly I don't have it installed. I will try to get it and see if the behaviour changes with the compiler version early next week.

amgc500:
Thanks for looking so quickly!

Yes, gfortran version 12.2.0 on Mac OS Ventura

valeriaRaffuzzi:
This was addressed in Mikolaj's pull request and should now be fixed.

PR #27, opened 2023-02-13
Fix uninitialised flag in json printer

Mikolaj-A-Kowalski:
A missing initialisation have been causing segmentation error when using JSON output format. It was revealed with gfortran 12 (#26) . For earlier compilers it must have worked due to luck I think.


valeriaRaffuzzi (approved):
Sounds good, thanks for looking into it!

PR #32, opened 2023-03-28
Adding check for undefined material in transport operators

valeriaRaffuzzi:
Adding check for undefined material in transport operators, useful to avoid segmentation faults. 

An undefined material mostly happens if there is an error in the geometry input.


ChasingNeutrons (approved):
This looks very good to me. One thing I might suggest adding to the error message is the particle location! Might make debugging the input file one step simpler.

valeriaRaffuzzi:
I agree, good idea

PR #39, opened 2023-05-29
Revert "Var reduction"

valeriaRaffuzzi:
Reverts CambridgeNuclear/SCONE#2

/home/vr339/SCONE-dump/SCONE/Geometry/Fields/VectorFields/weightWindowsField_class.f90:13:6:

   13 |   use tallyMapFactory_func,       only : new_tallyMap
      |      1
Fatal Error: Cannot open module file ‘tallymapfactory_func.mod’ for reading at (1): No such file or directory
compilation terminated.
make[2]: *** [CMakeFiles/scone.dir/build.make:1896: CMakeFiles/scone.dir/Geometry/Fields/VectorFields/weightWindowsField_class.f90.o] Error 1
make[2]: *** Waiting for unfinished jobs....
[ 41%] Building Fortran object CMakeFiles/scone.dir/NuclearData/emissionENDF/angleLawENDF/angleLawENDFslot_class.f90.o
[ 42%] Building Fortran object CMakeFiles/scone.dir/Tallies/TallyResponses/tallyResponseSlot_class.f90.o
make[1]: *** [CMakeFiles/Makefile2:990: CMakeFiles/scone.dir/all] Error 2
make: *** [Makefile:95: all] Error 2

Nasty error when compiling weightWindowsField_class.f90. This comes from trying to import tallyMapFactor_func, even though that line is used in multiple other parts of the code. By commenting out 
use tallyMapFactory_func,       only : new_tallyMap
and the line where new_tallyMap is used, all is fine...... ideas?


Mikolaj-A-Kowalski:
Just before the error CMake seems to claim to drop dependency on tallymapfactory due to a circular dependency.  This is why the error appears. 
That being said it is really not clear to me where the circular dependency is supposed to be...

PR #41, opened 2023-06-02
Fix Continious Integration workflows

Mikolaj-A-Kowalski:
Not ready to merge yet. 

Will collect fixes to the GitHub Actions etc.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -1,8 +1,10 @@
 name: build-and-test-ubuntu
-on: 
-  pull_request: 
-    branches:
-      - main
+on:
+  workflow_dispatch:
+  push:
+    branches: [ "main" ]
+  pull_request:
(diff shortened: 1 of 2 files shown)


Mikolaj-A-Kowalski:
So I think it should be ready to marge. It fixes the mistake with the incorrect path to status badge that I have made and (should?)* add the manual trigger for the workflow. 

Still we only compile with a single compiler. We should extend to all available `gfortran`s on Ubuntu 20.04LTS, but I think it will be better to do it as a separate PR.   

* It does not work on my branch but than it is not the default branch and documentation seems to claim it needs to be one: 
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

valeriaRaffuzzi (approved):
Looks good, thanks a lot for fixing this!! 

Go ahead and merge!

Issue #42, opened 2023-06-05
Test diffrent compilers in CI


Mikolaj-A-Kowalski:
Currently the `build-and-test` workflow only compiles SCONE with the default Fortran compiler of Ubuntu 20.04LTS (gfortran-11). However we would probably wish to test with more versions to see if there are any problems. In particular: 
- gfortran-8 (which will become the lowest supported version soon
- gfortran-12 (or whichever is the latest version at the moment) 

For good measure we could just cover them all for `>8`. 

Basically I think we should aim to have a similar solution we had on BitBucket, with containers having the right python for pfUnit, compiler and pfUnit preinstalled (to save build time). 

I will try to set it up soon.

Mikolaj-A-Kowalski:
Resolved by #45

PR #45, opened 2023-06-09
Add diffrent compilers to CI workflow

Mikolaj-A-Kowalski:
Addresses Issue #42. 

It switches the build strategy to use [Docker containers](https://hub.docker.com/r/mikolajkowalski/scone-test) which are build and published using a GitHub Actions workflow from another repository. I set it up as private for now, but perhaps it may be good idea to make it open. 

I checked that new build-and-test workflow runs fine on my fork, so I do not expect any issues on main. 

GFortran 11 and 12 is not tested because it is not trivial to get a Linux system that has it as a package, while the Python version is old enough to run with pFUnit in version 3.3.3. This makes #10 more and more paramount...

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -9,31 +9,11 @@ on:
 jobs:
   build-and-test:
     runs-on: ubuntu-20.04
+    container:
+      image: mikolajkowalski/scone-test:gfortran10
     steps:
       - uses: actions/checkout@v3
-      # We need to use older Python due to PfUnit 3
-      - uses: actions/setup-python@v4
-        with:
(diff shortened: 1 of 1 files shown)


Mikolaj-A-Kowalski:
The pFUnit is now simply pre-installed on the Docker containers. I hoped that it would help everything to run faster, but it seems that it is not really the case :-/ I suppose inside the docker container only one CPU of the Ubuntu host is seen(?).

valeriaRaffuzzi:
I see. It was worth a try! It can be something to fix in the future if there's a way? But nothing urgent.

valeriaRaffuzzi (approved):

Issue #38, opened 2023-05-24
Reducing output file memory


valeriaRaffuzzi:
At the moment, when writing the output file, all the tallies print their results in a unique charTape instance, and this gets written in the output file at once.

This approach is memory expensive when there's a lot of tallies, since the memory taken by the tallies gets duplicated when the results are written in the charTape. It would be more memory efficient to write the charTape into the output file after each tally is printed, and then clear the charTape. This would avoid long memory-intensive tapes.

PR #46, opened 2023-07-04
Stream output files instead of writing them in-memory first 

Mikolaj-A-Kowalski:
This resolves Issue #38.  However, I cannot help but to think it is a bit filthy... that being said I am not sure how exactly to make it nicer. 

The problem with streaming to the output file directly is that it is convenient to have a 'backtrace' capability, when writing the printers for different output formats. This allows, for example, to remove trailing commas with relative ease (crucial for JSON). 

To 'have a cake and eat it' this implements a `delayedStream` which works a bit like a buffered output, but when the buffer is flushed it retains `MIN_SIZE=16` last characters [as opposed to printing everything to the disk]. 

What I don't like about the changes is the way the file ownership is handled. It is made to be a responsibility of the `outputFile` class, which opens a file in `init`and closes it in `finalisation` (which is also the destructor). However, the printing is implemented by the `asciiOutput_inter` which owns an instance of `delayedStream`. The appropriate output unit is set by a subroutine.  

The other thing I fear may be confusing is that if one does not supply `filename` to `outputFile` no file will be produced (this is required for unit testing). 

In short this PR may still require some iterations [in addition to squashing the fixups!]

### DIFF
--- a/UserInterface/fileOutput/asciiJSON_class.f90
+++ b/UserInterface/fileOutput/asciiJSON_class.f90
@@ -96,7 +96,7 @@ contains
   end subroutine endFile
 
   !!
-  !! Return approperiate extension for the file
+  !! Return appropriate extension for the file
   !!
   !! See asciiOutput_inter for details
   !!
--- a/UserInterface/fileOutput/asciiJSON_class.f90
+++ b/UserInterface/fileOutput/asciiJSON_class.f90
@@ -21,7 +21,7 @@ module asciiJSON_class
   !! There is indentation for each block and entry, but long multi-dimensional (N-D) arrays
   !! are printed on a single line. Also N-D arrays are represented as a JSON array of arrays.
   !!
-  !! JSON output is fairly standard so it can be easlily read into Python and other enviroments
+  !! JSON output is fairly standard so it can be easily read into Python and other environments
   !!
   !! TODO:
   !!  Currently there is a dirty fix to remove commas at the end of a block, which requires to
@@ -39,7 +39,6 @@ module asciiJSON_class
   !!
   type, public, extends(asciiOutput) :: asciiJSON
     private
-    type(charTape)    :: output
     integer(shortInt) :: ind_lvl = 0
 
     integer(shortInt), dimension(:), allocatable :: shapeBuffer
@@ -49,8 +48,8 @@ module asciiJSON_class
 
   contains
     procedure :: init
+    procedure :: endFile
     procedure :: extension
-    procedure :: writeToFile
 
     procedure :: startBlock
     procedure :: endBlock
@@ -71,10 +70,30 @@ contains
     class(asciiJSON), intent(inout) :: self
 
     ! Add the initial bracket
-    call self % output % append( "{" // NEWLINE)
+    call self % append( "{" // NEWLINE)
     self % ind_lvl = 1
 
   end subroutine init
+
+  !!
+  !! Finalise the printer
+  !!
+  !! See asciiOutput_inter for details
+  !!
+  subroutine endFile(self)
+    class(asciiJSON), intent(inout) :: self
+
+    ! Dirty fix
+    ! Remove comma from the last entry by rewind
+    ! Need to check that the character is a comma to avoid removing { when there is an empty block
+    ! TODO: Find better way. Will clash with any proper stream output
+    if (self % peek(2) == ",") then
+      call self % cut(2)
+      call self % append(NEWLINE)
+    end if
+    call self % append("}" // NEWLINE)
+
+  end subroutine endFile
 
   !!
   !! Return approperiate extension for the file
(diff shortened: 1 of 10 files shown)


UserInterface/fileOutput/asciiJSON_class.f90
```diff
@@ -71,51 +70,43 @@ subroutine init(self)
...
     end if
+    call self % append("}" // NEWLINE)
 
-    ! Close the file
-    call self % output % append("}")
+  end subroutine endFile
 
-    form = '(A' // numToChar(self % output % length()) // ')'
+  !!
+  !! Return approperiate extension for the file
```
valeriaRaffuzzi, line 99:
```suggestion
  !! Return appropriate extension for the file
```

UserInterface/fileOutput/asciiMATLAB_class.f90
```diff
@@ -90,33 +87,29 @@ subroutine init(self)
...
 
-    str = 'm'
+    ! Nothing to do
 
-  end function extension
+  end subroutine endFile
 
   !!
-  !! Print the output to the given unit
+  !! Return approperiate extension for the file
```
valeriaRaffuzzi, line 102:
```suggestion
  !! Return appropriate extension for the file
```

UserInterface/fileOutput/dummyPrinter_class.f90
```diff
@@ -23,6 +23,7 @@ module dummyPrinter_class
 
   contains
     procedure :: init
+    procedure :: endFile
     procedure :: extension
     procedure :: writeToFile
```
valeriaRaffuzzi, line 28: I thought writeToFile was gone in the asciiOutput class?

UserInterface/fileOutput/outputFile_class.f90
```diff
@@ -210,17 +211,54 @@ module outputFile_class
...
     allocate( self % output, source = new_asciiOutput(self % type))
 
+    ! Open file if given
+    ! Otherwise make sure that the unit is not opened
+    if (present(filename)) then
+      ! Because GFotran hates implicit allocation
+      ! It causes buggy warnings to appear in compilation
+      allocate(self % outputFileName, source = trim(filename) // "." // self % output % extension())
+
+      open(newunit = self % outputUnit, &
```
valeriaRaffuzzi, line 240: It looks like 'error' is gone from this list.

valeriaRaffuzzi: same with errorMsg

UserInterface/fileOutput/outputFile_class.f90
```diff
@@ -233,58 +271,32 @@ subroutine init(self, type, fatalErrors)
...
-    ! Catch error
-    if (error /= 0 ) call fatalError(Here, errorMsg)
-
-    ! Write to file
-    call self % output % writeToFile(unitNum)
-
-    ! Close file
-    close(unitNum)
+  !! NOTE:
+  !!   This subroutine WILL NOT be called if the `end program` statement is
```
valeriaRaffuzzi, line 277: This comment is not clear.

UserInterface/fileOutput/outputFile_class.f90
Mikolaj-A-Kowalski: :fearful: Ups... not great thing to forget

UserInterface/fileOutput/dummyPrinter_class.f90
Mikolaj-A-Kowalski: It is... I think this is an oversight that this got left in.

UserInterface/fileOutput/outputFile_class.f90
Mikolaj-A-Kowalski:
```Fortran
  !!  NOTE:
  !!    Being marked `final`(a deconstructor)  this subroutine will be normally implicitly called by
  !!    by the compiler whenever a variable of `type(outputFile)` goes out of scope (e.g. when program 
  !!    execution reaches `end subroutine`, `end function` or `end block` statement. However, the rule in 
  !!    Fortran is that the decostructors are NOT called at the end of the program `end program`. In that case 
  !!    if one uses the output file as a module variable or defines it inside the program block, the `finalisation` 
  !!    must be called explicitly to ensure or data from the buffer is written to the disk 
  !!     
```
Does this work better?

valeriaRaffuzzi (changes requested):
See comments.

UserInterface/fileOutput/dummyPrinter_class.f90
Mikolaj-A-Kowalski: Should be removed now.

UserInterface/fileOutput/outputFile_class.f90
Mikolaj-A-Kowalski: Should be fixed.

UserInterface/fileOutput/outputFile_class.f90
Mikolaj-A-Kowalski: Comment has been changed.

Mikolaj-A-Kowalski:
So I think I have addressed all the issues now... 
If it looks ok I will rebase on the current main (and squash all the fixups!) and then it should be ready to be merged.

docs/Output.rst
```diff
@@ -76,9 +77,10 @@ Writing to an output file in SCONE is done through a series of calls to appropri
   !! Same goes for blocks
   call out % endBlock()
 
-  !! When done, we can write output to a file
-  !! Extension will be determined by the format printer
-  call out % writeToFile("./path/to/output/without_any_extension")
+  !! We need to make sure that output file is properly finalised. We can skip
+  !! this line if we we are not in a `program` block e.g. in `subroutine` or `function`
```
valeriaRaffuzzi, line 81:
```suggestion
  !! this line if we are not in a `program` block e.g. in `subroutine` or `function`
```
Typo!

valeriaRaffuzzi (approved):
Very nice, I think this all looks great and it's a much better alternative to write the output file than keep it all in memory! 

I think there's just a small leftover typo that I spotted in the docs (soon SCONE will be typo free!!!) but then feel free to rebase and merge

PR #49, opened 2023-07-05
Add debug and a build with no openmp to the CI

### LINKED ISSUE #48: Add compilation with diffrent options to CI
As a part of continuous integration, the build with `-DOPENMP=OFF` and `-DDEBUG=ON` should be performed as well.

Mikolaj-A-Kowalski:
Adds to additional build and test jobs that verify if SCONE compiles and passes tests with `DEBUG` option set to ON as well as with OpenMP switched off. 

Only a single compiler for each case is used. I don't know if it might be worth to extend it to all compilers. I am a bit paranoid about the usage of the GitHub Actions resources.  

P.S. Resolves #48

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -24,3 +24,31 @@ jobs:
           make -j
           make test
           cd -
+  build-and-test-debug:
+    runs-on: ubuntu-20.04
+    container:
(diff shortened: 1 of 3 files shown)


Mikolaj-A-Kowalski:
True. We can monitor how our usage will look like :crossed_fingers:.

PR #50, opened 2023-07-05
Fix one thread OMP reproducibility

Mikolaj-A-Kowalski:
I noted that reproducibility seemed to be broken when OpenMP is enabled. While it is expected for runs with multiple threads, it was happening for a single thread as well. 

As expected it turned out to be a typo. The `pRand` variable was being used uninitialised as a result of the type in the set-up of directives. 

I guess the only question is if we wish to integrate the fix quickly or push this PR to have a full OpenMP reproducibility (regardless of the number of threads)?

### DIFF
--- a/ParticleObjects/Source/source_inter.f90
+++ b/ParticleObjects/Source/source_inter.f90
@@ -100,8 +100,9 @@ contains
       type(particleDungeon), intent(inout) :: dungeon
       integer(shortInt), intent(in)        :: n
       class(RNG), intent(in)               :: rand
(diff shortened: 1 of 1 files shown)


ChasingNeutrons (approved):
Can't believe we let this one slip! Thanks for finding it. I think maybe best to open an issue for full reproducibility and get this one now.

Mikolaj-A-Kowalski:
I think you are correct. I will merge this one right now and open an issue.

PR #53, opened 2023-07-09
Spectrum of MG particles with point source

ChasingNeutrons:
Allow MG particles to have an energy distribution. Also typo fix in fissionSource, bug fix in pointSource, and moved nG from baseMgNeutronDatabase to the superclass

### DIFF
--- a/NuclearData/mgNeutronData/baseMgNeutron/baseMgNeutronDatabase_class.f90
+++ b/NuclearData/mgNeutronData/baseMgNeutron/baseMgNeutronDatabase_class.f90
@@ -45,7 +45,6 @@ module baseMgNeutronDatabase_class
   !!   mats       -> array containing all defined materials (by matIdx)
   !!   majorant   -> majorant xs for delta tracking
   !!   activeMats -> list of matIdxs of materials active in the problem
-  !!   nG         -> number of energy groups
   !!
   !! Interface:
   !!   nuclearDatabase interface
@@ -54,7 +53,6 @@ module baseMgNeutronDatabase_class
     type(baseMgNeutronMaterial), dimension(:), pointer :: mats => null()
     real(defReal), dimension(:), allocatable           :: majorant
     integer(shortInt), dimension(:), allocatable       :: activeMats
-    integer(shortInt)                                  :: nG = 0
 
   contains
     ! Superclass Interface
--- a/NuclearData/mgNeutronData/mgNeutronDatabase_inter.f90
+++ b/NuclearData/mgNeutronData/mgNeutronDatabase_inter.f90
@@ -1,5 +1,7 @@
 module mgNeutronDatabase_inter
 
+  use numPrecision
+
   ! Nuclear Data Interfaces & Objects
   use nuclearDatabase_inter, only : nuclearDatabase
 
@@ -16,9 +18,13 @@ module mgNeutronDatabase_inter
   !!
   !! It does nothing, It adds nothing,
   !! It just provides a common superclass for related classes
+  !! and holds the number of energy groups
+  !!
+  !! Public members: 
+  !!   nG -> number of energy groups
   !!
   type, public, abstract, extends(nuclearDatabase) :: mgNeutronDatabase
-
+    integer(shortInt)  :: nG = 0
   end type mgNeutronDatabase
 
 contains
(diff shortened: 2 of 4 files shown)


ParticleObjects/Source/pointSource_class.f90
```diff
@@ -210,9 +226,16 @@ subroutine sampleEnergy(self, p, rand)
     class(pointSource), intent(inout)   :: self
     class(particleState), intent(inout) :: p
     class(RNG), intent(inout)           :: rand
+    real(defReal)                       :: r
+    integer(shortInt)                   :: g
 
     if (self % isMG) then
-      p % G = self % G
+      r = rand % get()
```
Mikolaj-A-Kowalski, line 233:
```suggestion
      !! TODO: Replace this single, common procedure to sample discrete distributions
      r = rand % get()
```
Sadly the Alias method is implemented... but in the infamous photon branch... :-(

NuclearData/mgNeutronData/baseMgNeutron/baseMgNeutronDatabase_class.f90
```diff
@@ -45,7 +45,7 @@ module baseMgNeutronDatabase_class
   !!   mats       -> array containing all defined materials (by matIdx)
   !!   majorant   -> majorant xs for delta tracking
   !!   activeMats -> list of matIdxs of materials active in the problem
-  !!   nG         -> number of energy groups
+  !!   nG         -> number of energy groups (from superclass)
```
Mikolaj-A-Kowalski, line 48:
```suggestion
```
Maybe we should move this to `mgNeutronDatabase`. (add there `Public Members:` section with nG?

NuclearData/mgNeutronData/mgNeutronDatabase_inter.f90
```diff
@@ -16,9 +18,10 @@ module mgNeutronDatabase_inter
   !!
   !! It does nothing, It adds nothing,
   !! It just provides a common superclass for related classes
+  !! and holds the number of energy groups
```
Mikolaj-A-Kowalski, line 21:
```suggestion
  !! and holds the number of energy groups
  !!
  !! Public members: 
  !!   nG -> number of energy groups
```

Mikolaj-A-Kowalski (approved):
I see no reason not to merge after the few cosmetic changes. 
Thanks for the development!

ParticleObjects/Source/pointSource_class.f90
```diff
@@ -131,19 +136,30 @@ subroutine init(self, dict, geom)
...
+      nucData => ndReg_getNeutronMG()
+      if(.not.associated(nucData)) call fatalError(Here, 'Failed to retrieve Nuclear Database')
+
+      if (size(temp) /= nucData % nG) then
+        call fatalError(Here, 'Source energy group distribution must have '//numToChar(nucData % nG)//' groups')
+      end if
+
+      allocate(self % probG(nucData % nG))
+      self % probG = temp
+      self % probG = self % probG / self % probG
```
valeriaRaffuzzi, line 157: Should it be divided by the sum?

valeriaRaffuzzi:
What about leaving the option to use only 'G' as well? Sometimes in fixed source one could have lots of groups (vitamin-j is 175), and I've definitely used up to 700. If we want to do fusion, where particles are mono-energetic, it's definitely convenient to leave the option!

ParticleObjects/Source/pointSource_class.f90
ChasingNeutrons: Thank you for spotting that. I could swear I did it... Need to write unit tests for sources :/

Mikolaj-A-Kowalski:
Also perhaps we could squash all the commits into one (or few) with `git rebase -i main` (and selecting `squash` [`s`] option for commits that come from the GitHub edits etc...?

ChasingNeutrons:
Ok, after making a slight mess that didn't compile, I think those things should be fixed and now there should be a convenient option for monoenergetic MG particles.

ChasingNeutrons:
Did the squash work as intended?

Mikolaj-A-Kowalski:
Not yet! You may still need to force push your modified branch. (`git push -f`). 
But before doing that it is always good idea to make sure that some backup of the branch being overridden exists somewhere (just in case...) 
Thanks for the rebase! Sorry for being a pain

ChasingNeutrons:
I don't entirely understand what happened but I _think_ that should be it.

Mikolaj-A-Kowalski:
Yup! It is done. Shell I merge?

PR #54, opened 2023-07-14
Nth collision tally

valeriaRaffuzzi:
This PR adds a 1D tally map for the number of collisions that a particle went through. 

A collision counter is added to the particle. This is incremented in collisionProcessor_inter, after a collision happened and before reportOutCollision. When a particle is generated by fission, the counter has to be set to zero for the new source particles.


ChasingNeutrons (commented):
This all looks good.
My main comment is that, I think, for practical applications, we are not often interested in the 15th collided neutrons, for example, but neutrons which have undergone some range of collisions - maybe the uncollided and the more than once collided for example. Otherwise this would require knowing how many collisions a neutron will undergo to estimate. Are there any convenient ways in which we can tally these? Would this need a new integer map type? Is this new map type desirable?

ChasingNeutrons (approved):

ParticleObjects/particle_class.f90
ChasingNeutrons, line 1: I think there should be a check on collisionN here too.

valeriaRaffuzzi: Done now!

PR #56, opened 2023-07-21
Update the JEZ example input file

Mikolaj-A-Kowalski:
The JEZ as we have right now will not run with modern nuclear data evaluations (e.g. JEFF 3.3 from NEA), since they do not provide the elemental Gallium data (only individual isotopes)  

Use the specification from [LA-UR-14-21554 report](https://www.osti.gov/biblio/1122897).
Makes the example compatible with newer nuclear data evaluations which do not provide the elemental gallium.

New JEZ does not stay k_eff=1.0 for JEFF 3.1 but is spot on for JEFF 3.3.

### DIFF
--- a/InputFiles/JEZ
+++ b/InputFiles/JEZ
@@ -1,3 +1,8 @@
+//
+// PU-MET-FAST-001 rev. 3
+//
+// As described in LA-UR-14-21554
+//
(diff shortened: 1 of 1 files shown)


Mikolaj-A-Kowalski:
Thanks. I will merge!

PR #57, opened 2023-07-21
Remove `vector_class`

Mikolaj-A-Kowalski:
It is unused for a long time. It was an oversight it has survived that long.


valeriaRaffuzzi (approved):
Nice cleaning up!

PR #58, opened 2023-07-23
Fix to mgXsClerk

ChasingNeutrons:
There was an error in the mg clerk which would fail to recognise when an entry was out of bounds. This was due to having an incorrect 'exit' condition on the tally - naturally it would be a check for a map index of 0, but because of flipping the energy groups around (for postprocessing convenience) this should actually be a check for nG + 1. This now runs and unit tests pass.


valeriaRaffuzzi (approved):
Thanks for spotting this! It was my bad, this slipped when I reversed the order of the energy groups to follow the fast -> thermal convention.

PR #59, opened 2023-07-27
Update mgXsClerk_class.f90

ChasingNeutrons:
Fix to prevent seg fault when no energy map is given. Adds a check for whether the energyMap is allocated and behaves accordingly.

### DIFF
--- a/Tallies/TallyClerks/mgXsClerk_class.f90
+++ b/Tallies/TallyClerks/mgXsClerk_class.f90
@@ -739,7 +739,7 @@ contains
     end if
 
     ! Print energy map information
-    if (allocated(self % energyMap)) self % energyMap % print(outFile)
+    if (allocated(self % energyMap)) call self % energyMap % print(outFile)
     resArrayShape(1) = self % energyN
 
     ! If a space map print map information
--- a/Tallies/TallyClerks/mgXsClerk_class.f90
+++ b/Tallies/TallyClerks/mgXsClerk_class.f90
@@ -739,12 +739,8 @@ contains
     end if
 
     ! Print energy map information
-    if (allocated(self % energyMap)) then
-      call self % energyMap % print(outFile)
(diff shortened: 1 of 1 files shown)


valeriaRaffuzzi (commented):
You could do: 

if (allocated(self % energyMap)) call self % energyMap % print(outFile)
resArrayShape(1) = self % energyN

It's a bit cleaner.

Tallies/TallyClerks/mgXsClerk_class.f90
```diff
@@ -739,8 +739,12 @@ subroutine print(self, outFile, mem)
...
 
     ! Print energy map information
-    call self % energyMap % print(outFile)
-    resArrayShape(1) = self % energyN
+    if (allocated(self % energyMap)) then
+      call self % energyMap % print(outFile)
+      resArrayShape(1) = self % energyN
+    else
+      resArrayShape(1) = 1
+    end if
```
valeriaRaffuzzi, line 747:
```suggestion
    if (allocated(self % energyMap)) self % energyMap % print(outFile)
    resArrayShape(1) = self % energyN
```
This makes it more compact!

valeriaRaffuzzi (changes requested):

Tallies/TallyClerks/mgXsClerk_class.f90
```diff
@@ -739,7 +739,7 @@ subroutine print(self, outFile, mem)
     end if
 
     ! Print energy map information
-    call self % energyMap % print(outFile)
+    if (allocated(self % energyMap)) self % energyMap % print(outFile)
```
valeriaRaffuzzi, line 742:
```suggestion
    if (allocated(self % energyMap)) call self % energyMap % print(outFile)
```
My bad, I forgot the call!

valeriaRaffuzzi (approved):

PR #63, opened 2023-08-04
DBRC

valeriaRaffuzzi:
Adds routines to perform target velocity sampling with DBRC, and routines to calculate the scattering temperature majorant cross section in the aceNuclearDatabase.

I still need to check whether this implementation breaks when trying to use a different type of NuclearDatabase (e.g., aceNuclearDatabaseUni), and possibly add tests for some of the DB functions.


valeriaRaffuzzi:
A question about this implementation: in the collisionProcessors, right now aceNeutronDB and aceNeutronNuclide are casted to get cross sections and info about DBRC. This is now working fine because we removed the unionised databases and nuclides. However, for the future, it might be a good investment to move some DBRC functions (the interface) to ceNeutronDB and ceNeutronNuclide, in case we add new databases, so that the collisionProcessor can communicate with ceNeutronDB?

Mikolaj-A-Kowalski:
I wholeheartedly agree. Certainly the `hasDBRC` function feels like it would fit in the `ceNuclide`. A part of me is just wondering if it shouldn't be renamed to something more generic like `alwaysMovingTarget`. Basically it comes down to if there are other cases apart from DBRC when we might wish to force that a nuclide will be processed a moving target.... at the moment I don't know. 

As for the database we will need to change it, but I think it must be a part of a more major refactor of nuclear data in general. Basically we need to somehow allow the client code to choose a temperature at which they want cross-sections. Then getting the value at 0K should be OK. 

I don't know if you agree with that?

valeriaRaffuzzi:
> I wholeheartedly agree. Certainly the `hasDBRC` function feels like it would fit in the `ceNuclide`. A part of me is just wondering if it shouldn't be renamed to something more generic like `alwaysMovingTarget`. Basically it comes down to if there are other cases apart from DBRC when we might wish to force that a nuclide will be processed a moving target.... at the moment I don't know.
> 

as things are now, in the collisionProcessor the if the flag hasDBRC is true this will lead to using DBRC, rather than something generic. So I don't mind calling the flag hasDBRC for now, and if some future different implementation comes up that we can always make it more generic!

> As for the database we will need to change it, but I think it must be a part of a more major refactor of nuclear data in general. Basically we need to somehow allow the client code to choose a temperature at which they want cross-sections. Then getting the value at 0K should be OK.
> 

I'm not sure I understand this point. For now I was just talking about moving the function to get the elastic majorant to ceNeutronDB (and make it deferred), so that the collisionProcessor wouldn't need to call aceNeutronDB directly.

Mikolaj-A-Kowalski:
> > As for the database we will need to change it, but I think it must be a part of a more major refactor of nuclear data in general. Basically we need to somehow allow the client code to choose a temperature at which they want cross-sections. Then getting the value at 0K should be OK.
> 
> I'm not sure I understand this point. For now I was just talking about moving the function to get the elastic majorant to ceNeutronDB (and make it deferred), so that the collisionProcessor wouldn't need to call aceNeutronDB directly.

Sorry I got confused about the point. 
I think you are right that this function could live in the `ceNeutronDB`. Also a similar procedure will be required in TMS right (but with total XS instead of just elastic)?

valeriaRaffuzzi:
> Sorry, I got confused about the point :-) I think you are right that this function could live in the `ceNeutronDB`. Also a similar procedure will be required in TMS right (but with total XS instead of just elastic)?

Yes, I think maybe what you were saying refers more to TMS. 

I think so, the temperature majorant should be needed for TMS too. When we get TMS, we just need to make sure the implementations of TMS and DBRC follow the same logic, which shouldn't be too challenging!

valeriaRaffuzzi:
This should be all! I tested it against Serpent one last time and it agrees, so I think this is it.
I also updated the manual.

CollisionOperator/CollisionProcessors/neutronCEimp_class.f90
```diff
@@ -644,27 +657,53 @@ subroutine scatterFromMoving(self, p, collDat, reac)
...
 
-    ! Sample velocity of target
-    V_t = targetVelocity_constXS(p % E, dir_pre, A, kT, p % pRNG)
+    ! Sample target velocity with constant XS or with DBRC
+    ! Check energy range
+    eRange = ((p % E <= self % DBRCeMax) .and. (self % DBRCeMin <= p % E))
+    ! Check if DBRC is on for this target nuclide
+    hasDBRC = self % nuc % hasDBRC()
+
+    if (eRange .and. hasDBRC) then
```
Mikolaj-A-Kowalski, line 687: With regards to the previous comment here would be `.not. inEnergyRange`

CollisionOperator/CollisionProcessors/neutronCEimp_class.f90
```diff
@@ -644,27 +657,53 @@ subroutine scatterFromMoving(self, p, collDat, reac)
...
+
+      ! Retrieve 0K nuclide index from DBRC nuclide map
+      nucIdx = self % xsData % mapDBRCnuc % get(nucIdx)
+
+      ! Reassign pointer for the 0K nuclide
+      self % nuc => ceNeutronNuclide_CptrCast(self % xsData % getNuclide(nucIdx))
```
Mikolaj-A-Kowalski, line 693: I am wondering if it wouldn't be petter to have a specific, local pointer (local variable) for the 0K version of the nuclide? 
Basically right now the nuclide which is saved in the processor is being changed if DBRC happens to be applied and I fear that it may cause confusion at some point in the future if somebody tries to access the nuclide to e.g. apply cut-offs or something similar?

CollisionOperator/CollisionProcessors/neutronCEstd_class.f90
```diff
@@ -404,24 +416,53 @@ subroutine scatterFromMoving(self, p, collDat, reac)
...
+      maj = self % xsData % getScattMicroMajXS(p % E, kT, A, nucIdx)
+
+      ! Use DBRC to sample target velocity
+      V_t = targetVelocity_DBRCXS(self % nuc, p % E, dir_pre, A, kT, p % pRNG, maj)
+
+    else
+      ! Constant cross section approximation
+      V_t = targetVelocity_constXS(p % E, dir_pre, A, kT, p % pRNG)
+
+    end if
```
Mikolaj-A-Kowalski, line 465: Sadly due to code duplication :-( the same questions as in the `neutronCEimp` version apply

CollisionOperator/scatteringKernels_func.f90
```diff
@@ -149,6 +152,89 @@ function targetVelocity_constXS(E,dir,A,kT,rand) result (V_t)
...
+      P_acc = rel_v / (Y + X)
+
+      ! Accept or reject mu and target velocity
+      if (P_acc < r3) cycle rejectionLoop
+
+      ! Relative energy = relative velocity **2 due to sqrt(Mn/2) scaling factor
+      rel_E = (rel_v**2 * kT / A)
+
+      ! Find 0K scattering xs of target at relative energy
+      xs_rel_v = nuc % elScatteringXS(rel_E)
```
Mikolaj-A-Kowalski, line 215: Perhaps it would be a good idea to move the common part of the sampling between DBRC and non-DBRC target velocity sampling to a shared, separate procedure (private for the module)? It would avoid some code duplication.

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
```diff
@@ -645,6 +707,68 @@ subroutine init_urr(self)
 
   end subroutine init_urr
 
+  !!
+  !!  Checks through all nuclides, creates map with nuclides present and corresponding 0K nuclide
+  !!
+  !!  NOTE: compares the first 5 letters of the ZAID.TT. It would be wrong with isotopes
+  !!        with Z > 99
```
Mikolaj-A-Kowalski, line 714: I am a bit of scared of this one since some nuclides have only 4 digits in ZAID (e.g. `7014`) not that they will ever be significant for DBRC but if they are accidentally included in the list there is a chance code might do some weird things. 

Perhaps we can use the `.` as the separator and just compare the ZAID part before the first `.`?

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
```diff
@@ -645,6 +707,68 @@ subroutine init_urr(self)
...
+        ! Exit loop if a match is found
+        if (nucSet % atKey(k) == nuc0K) then
+          idx = nucSet % atVal(k)
+          exit
+        end if
+
+        ! Increment index
+        k = nucSet % next(k)
+
+      end do
```
Mikolaj-A-Kowalski, line 746: I think this linear search could be replaced with a get procedure, right? 
We just want to find the value at the key which is the nuclide ZAID?

NuclearData/ceNeutronData/ceNeutronNuclide_inter.f90
```diff
@@ -262,6 +295,45 @@ pure function isFissile(self) result(isIt)
...
+  !!
+  !! Errors:
+  !!   None
+  !!
+  subroutine setDBRC(self)
+    class(ceNeutronNuclide), intent(inout) :: self
+
+    self % DBRC = .true.
+
+  end subroutine setDBRC
```
Mikolaj-A-Kowalski, line 315: Perhaps we should move it as an additional keyword in the `set` procedure. 
The original intention of having a single procedure that would be called with keyword, optional arguments was to simply the interface and avoid having to write may `set` procedures. I don't know if it ends up being convenient in practice though...

Mikolaj-A-Kowalski (commented):
Thank you for the PR! I am glad that the DBRC will be merged soon. 
I guess at some point we will have to think a bit about the collision processors and how they work... I am a bit uneasy about the code duplication there. But we will address it at a later time.

NuclearData/ceNeutronData/ceNeutronNuclide_inter.f90
valeriaRaffuzzi: I think it's not convenient with how things are implemented now, because setDBRC is called from aceNEutronDatabase and not in aceNeutronNuclide. So I would need to modify things a lot and maybe make them less neat (now we have a function called initDBRC in the aceNeutronDB, if I try to use set I would need to modify init and probably make it a bit messier). I don't think it's worth it.

Mikolaj-A-Kowalski: I don't know if it would really be much messier. I mean it would just boil down to changing the line 760 of `aceNeutronDatabase_class.f90` from:
```Fortran
call self % nuclides(nucSet % atVal(j)) % setDBRC()
```
to e.g.: 
```Fortran
call self % nuclides(nucSet % atVal(j)) % set(dbrc=.true.)
```
Basically there is no problem calling `set` multiple times in different places since if the keyword is not provided on invocation, the associated entry is simply unchanged.   
There is a good chance I am missing something but I have a trouble to understand what would necessitate a rewrite of the init procedures.

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
valeriaRaffuzzi: Yes, I don't know how this survived in there! Good point.

NuclearData/ceNeutronData/ceNeutronNuclide_inter.f90
valeriaRaffuzzi: Sure, I see! I think I hadn't realised that all the arguments of set were optional and one could just call it like that. Very convenient

CollisionOperator/scatteringKernels_func.f90
```diff
@@ -298,5 +264,53 @@ function sample_x3expx2(rand) result(sample)
...
+    ! Sample polar angle of target velocity wrt. neutron direction
+    mu = 2.0 * r2 - 1.0;
+
+    ! Calculate relative velocity between neutron and target
+    rel_v = sqrt(Y * Y + X * X - 2.0 * X * Y * mu)
+
+    ! Calculate Acceptance Propability
+    P_acc = rel_v / (Y + X)
+
+    ! Verigy acceptance condition
```
Mikolaj-A-Kowalski, line 306:
```suggestion
    ! Verify acceptance condition
```
:snake:

CollisionOperator/scatteringKernels_func.f90
```diff
@@ -298,5 +264,53 @@ function sample_x3expx2(rand) result(sample)
...
+    if (P_acc > r3) then
+      accept = .true.
+    else
+      accept = .false.
+    end if
```
Mikolaj-A-Kowalski, line 311:
```suggestion
    accept = P_acc > r3
```

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
```diff
@@ -713,51 +713,41 @@ end subroutine init_urr
...
-        k = nucSet % next(k)
+      ! Get ZAID with 0K temperature code
+      nuc0K = trim(nucDBRC(i))//'.00'
 
-      end do
+      ! Find the nucIdxs of the 0K DBRC nuclides
+      idx0K = nucSet % get(nuc0K)
 
-      ! Loop through nucSec again to find the nucIdxs of the DBRC nuclides with
+      ! Loop through nucSec to find the nucIdxs of the DBRC nuclides with
```
Mikolaj-A-Kowalski, line 735:
```suggestion
      ! Loop through nucSet to find the nucIdxs of the DBRC nuclides with
```

Mikolaj-A-Kowalski (approved):
Looks good! Just a couple of last few minor typos included as the suggestions :-) 
Thanks for the implementation!

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
```diff
@@ -645,6 +707,58 @@ subroutine init_urr(self)
 
   end subroutine init_urr
 
+  !!
+  !!  Checks through all nuclides, creates map with nuclides present and corresponding 0K nuclide
+  !!
+  !!  NOTE: compares the first 5 letters of the ZAID.TT. It would be wrong with isotopes
+  !!        with Z > 99
```
valeriaRaffuzzi, line 714:
```suggestion
```

PR #65, opened 2023-08-07
Remove all files about nuclearDB with unionised grids

valeriaRaffuzzi:
As discussed, this PR removes all the files related to the unionised grid nuclear databases and nuclides.


Mikolaj-A-Kowalski (commented):
As far as the modified files go it all seems all right. 
But probably we should also purge the remainders of the unionisation in the `ceNautronCache` as well.

valeriaRaffuzzi:
> As far as the modified files go it all seems all right. But probably we should also purge the remainders of the unionisation in the `ceNautronCache` as well.

Good point! The tests didn't complain so I forgot about it.

Mikolaj-A-Kowalski (approved):
Thanks for the changes! :seal: 
I keep thinking if there isn't any other place where trace of the unionised universes might remain. I think that both input files and the manual are clear. I cannot think of anything else.

valeriaRaffuzzi:
> Thanks for the changes! 🦭 I keep thinking if there isn't any other place where trace of the unionised universes might remain. I think that both input files and the manual are clear. I cannot think of anything else.

I tend to agree, I can't find anything else. If something else comes up otherwise we can fix it later on.

PR #66, opened 2023-08-11
Improve performance of the IO dictionary parser 

Mikolaj-A-Kowalski:
This PR makes two major changes that significantly improve the performance of dictionary parser 
- The format string when checking for a type of an entry is hardcoded, which avoids significant number of allocations 
- The reading from the file is changed to proceed in a stream fashion. This allow to remove the `MAX_COLUMN` restriction on the input files.

Previously increasing the value of `MAX_COLUMN` would make parsing extremely slow since the procedures searching for comments and TABS would traverse the entire line buffer (filled mostly with spaces). Thus the scaling was `O(N * MAX_COLUMN)` where `N` is number of lines in the input file. 

With this improvements the in/out transfers of large quantities of data should become more feasible.

### DIFF
--- a/DataStructures/dictParser_func.f90
+++ b/DataStructures/dictParser_func.f90
@@ -610,13 +610,15 @@ contains
   !! Interpret contents of buffer
   !!
   !! Sets self % type to {CONV_INT, CONV_FLOAT, CONV_CHAR}
-  !! loads value into approperite member {i, f, c}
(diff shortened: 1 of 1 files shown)


valeriaRaffuzzi:
Looks great, thanks a lot! It will definitely make my life much easier in the future :D

valeriaRaffuzzi (approved):

PR #68, opened 2023-08-14
Update mgXsClerk_class.f90

ChasingNeutrons:
Fix problem where inelastic S(a,b) was not scoring to mgXS tallies.


valeriaRaffuzzi (approved):
Thanks for spotting this!

PR #70, opened 2023-08-21
Use pFUnit 4 for testing

Mikolaj-A-Kowalski:
This aims to address #10 and switch the pFUnit to the version 4. Furthermore, it turns out that warnings were not enable when building tests. Consequently the warnings that are now raised are addressed.  

Note:
* Since we remove the custom `FindPFUnit.cmake` the hint enviromental variable has changed from `PFUNIT_INSTALL` to `PFUNIT_DIR`.
* The GitHub Actions will fail due to old pFUnit in the containers (they will need to be updated, I am not sure how to make it in a clean way)
*  pfUnit seems to have a bug that creates an unused variable when processing a Test Fixture. Sadly this means that the warning needs to be suppressed. 
* Tests use the same flags as SCONE.  Sadly it means that the optimisation is turned on which makes the tests compile much slower. 

In short term we should also aim to update the CMake configuration and use FetchContent to download the pFUnit automatically. I did not do this in this PR to keep it smaller. Not sure if this is a good idea or bad.  

P.S. Should definitely NOT be merged before the workshop since it will cause havoc...


Mikolaj-A-Kowalski:
@valeriaRaffuzzi We probably need to start to move this PR forward. 
I have updated the Github actions settings so the tests are run with images that come with pre-installed pFUnit 4.7.3, hence the automated tests pass now without any problem. 

The main issues to figure out at the moment is: 
1. Should we go over and do a review now or should I rebase on the current main before we do 
2. Do we accept the price of longer compilation time for the tests (now they run with optimisations enabled) or should I play with CMake settings a bit more to go back to compiling them with `-O0` 
3. Of course integrating this PR will cause havoc, but there is no way around it. What is the best way to proceed to minimise disruption to everybody.

valeriaRaffuzzi:
1. I don't have a strong preference, but I don't think it's necessary to rebase. 
2. How long are we talking about? The compilation is already quite slow, so it mainly depends on how tricky it is to change those flags. 
3. I don't think it will be a big deal. It's just about downloading the latest pFUnit, right? Maybe the most painless option would be to download automatically the right version of pFUnit as you were mentioning above, but I don't think it will be as bad as you say if we don't do that. 

Otherwise: If the name of the PFUNIT environmental variable has changed we should update the Installation doc as well.

Mikolaj-A-Kowalski:
1. Ok. We will just need to rebase before merge to make sure that all tests which were added since this branch diverged from `main` were updated to new syntax.  
2. Ouch :upside_down_face: ! And here I was thinking that SCONE compiles fast! :smiling_face_with_tear: New: 4m24s; Old 2m40s on my laptop with single core. 
3. It is mostly that all developments done before the this branch is merged will need to update the syntax of the test files and there is a good chance for many marge conflicts in tests!

DataStructures/Tests/charMap_test.f90
```diff
@@ -198,7 +196,7 @@ subroutine testLooping(this)
     character(nameLen)                        :: tKey
 
     ! Initialise parameters
-    KEYS_PAST = "This is not a Key. It's picture of a key!"
```
ChasingNeutrons, line 201: Why?!

ChasingNeutrons (approved):
Very very belatedly... Thanks for doing this! It looks like it was surprisingly straightforward, other than a lot of name changes.

Issue #69, opened 2023-08-21
Manual about Material


Congjin321:
I think the Material definition is not in a good position in the User Manual.

Mikolaj-A-Kowalski:
Hopefully the change in the manual is satisfactory :-) 
I will close the issue for now, but if you happen to think it is still confusing please feel free to reopen!

PR #71, opened 2023-08-25
Add section title for composition definition in the manual

Mikolaj-A-Kowalski:
Makes a small change to the manual to make the description of the material compositions easier to find (addresses #69 ). 
In the process also removes the trailing white spaces in the file (thus the size of the diff)


valeriaRaffuzzi (approved):
Looks good! Sorry for all the spaces, I must have left them when I split the lines

PR #73, opened 2023-08-25
Clean all text files to make them more UNIX

Mikolaj-A-Kowalski:
Basically applies large number of changes that does not change anything but just cleans up the files which do not conform to typical convention used in programming. That is files that:
* Have any trailing whitespace at the end of the line 
* Contain DOS-style line endings: CRLF 
* Do not terminate with the new line feed  

Sadly this is a very large change and may cause a significant number of annoying conflicts :-( 
I don't know if we wish to merge it now. Or if we want to include scripts that may make this clean-up (more or less) automated in the future.


valeriaRaffuzzi:
Other than the conflicts coming from the recent PRs that we just merged, I don't think that this PR will break anything. 

In the future probably having a script that automates this clean up would be great, maybe open an issue about it? Otherwise I think we can just resolve conflicts and merge without problems.

PR #74, opened 2023-08-29
Move reading of id, origin and rotation to `universe_inter`

Mikolaj-A-Kowalski:
Moves the code repeated in each universe subclass that sets up the properties of the base `universe` class to a separate, common function. 

This allows to avoid code repeat and will make the extension (to e.g. time-dependent transformations) easier.

### DIFF
--- a/Geometry/Universes/rootUniverse_class.f90
+++ b/Geometry/Universes/rootUniverse_class.f90
@@ -84,12 +84,11 @@ contains
     character(nameLen)                                        :: name
     character(100), parameter :: Here = 'init (rootUniverse_class.f90)'
 
-    ! Set ID
-    call dict % get(id, 'id')
-    if (id <= 0) call fatalError(Here, 'Universe ID must be +ve. Is: '//numToChar(id))
-    call self % setId(id)
+    ! Setup the base class
+    ! With: id, origin rotations...
+    call self % setupBase(dict)
 
-    ! Make shure root does not contain neither origin nor rotation
+    ! Make sure root does not contain neither origin nor rotation
     if (dict % isPresent('origin')) then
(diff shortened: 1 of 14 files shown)


Geometry/Universes/rootUniverse_class.f90
```diff
@@ -84,12 +84,11 @@ subroutine init(self, fill, dict, cells, surfs, mats)
...
-    call dict % get(id, 'id')
-    if (id <= 0) call fatalError(Here, 'Universe ID must be +ve. Is: '//numToChar(id))
-    call self % setId(id)
+    ! Setup the base class
+    ! With: id, origin rotations...
+    call self % setupBase(dict)
 
-    ! Make shure root does not contain neither origin nor rotation
+    ! Make sure root does not contain neither origin nor rotation
     if (dict % isPresent('origin')) then
```
valeriaRaffuzzi, line 92: dict % isPresent('origin') 
is repeated twice in line 92 and 95

valeriaRaffuzzi (changes requested):
It looks sensible to me, except for the one comment I left. Nice to avoid code repetition

Geometry/Universes/rootUniverse_class.f90
Mikolaj-A-Kowalski: Yikes!!! 
It has been there for long.... Good thing nobody was rotating root universes... :fearful:

Mikolaj-A-Kowalski: Should be resolved now.

valeriaRaffuzzi (approved):
Looks good now! At this rate SCONE will soon be typos free!

Issue #75, opened 2023-09-01
Code crashes if 'dist' in Surface tracking becomes NaN


anuj-git-coding:
Hello!

In case any modification in surface tracking transport operator leads to a 'NaN' flight distance, the code crashes without printing the root cause of the problem  (dist = 'Nan'). Perhaps a suitable error message will be helpful for users to debug the code?

Cheers
Anuj

Current crash message ->
Fatal has occurred in:
 decreaseLevel (coord_class.f90)

 Because:
 New nesting: 0 is invalid. Current nesting: 3
 <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
Note: The following floating-point exceptions are signalling: IEEE_INVALID_FLAG IEEE_UNDERFLOW_FLAG IEEE_DENORMAL
ERROR STOP

Error termination. Backtrace:
#0  0x7f64eee86d11 in ???
#1  0x7f64eee87859 in ???
#2  0x7f64eee88f87 in ???
#3  0x5587d6256cab in __errors_mod_MOD_fatalerror
        at /home/ad2221/SCONE_ORIGINAL/SCONE/SharedModules/errors_mod.f90:66
#4  0x5587d622400b in __coord_class_MOD_decreaselevel
        at /home/ad2221/SCONE_ORIGINAL/SCONE/Geometry/coord_class.f90:318
#5  0x5587d61e7037 in __geometrystd_class_MOD_move_nocache
        at /home/ad2221/SCONE_ORIGINAL/SCONE/Geometry/geometryStd_class.f90:238
#6  0x5587d621d73a in __transportoperatorst_class_MOD_surfacetracking
        at /home/ad2221/SCONE_ORIGINAL/SCONE/TransportOperator/transportOperatorST_class.f90:199
#7  0x5587d618d4b6 in __transportoperator_inter_MOD_transport.constprop.0
        at /home/ad2221/SCONE_ORIGINAL/SCONE/TransportOperator/transportOperator_inter.f90:111
#8  0x5587d6256f9a in __eigenphysicspackage_class_MOD_cycles._omp_fn.1
        at /home/ad2221/SCONE_ORIGINAL/SCONE/PhysicsPackages/eigenPhysicsPackage_class.f90:209
#9  0x7f64eead28e5 in ???
#10  0x5587d6257700 in __eigenphysicspackage_class_MOD_cycles
[...]

PR #87, opened 2023-09-14
Prevents NaN distances in transport operator

valeriaRaffuzzi:
Adresses issue #75

### DIFF
--- a/TransportOperator/transportOperatorDT_class.f90
+++ b/TransportOperator/transportOperatorDT_class.f90
@@ -49,11 +49,12 @@ contains
     ! Get majornat XS inverse: 1/Sigma_majorant
     majorant_inv = ONE / self % xsData % getMajorantXS(p)
 
-    ! Should never happen! Prevents NaN distances
-    if (majorant_inv /= majorant_inv) call fatalError(Here, "Majorant is 0")
+    ! Should never happen! Prevents Inf distances
+    if (abs(majorant_inv) > huge(majorant_inv)) call fatalError(Here, "Majorant is 0")
 
     DTLoop:do
       distance = -log( p% pRNG % get() ) * majorant_inv
+
 
       ! Move partice in the geometry
       call self % geom % teleport(p % coords, distance)
--- a/TransportOperator/transportOperatorDT_class.f90
+++ b/TransportOperator/transportOperatorDT_class.f90
@@ -49,6 +49,9 @@ contains
     ! Get majornat XS inverse: 1/Sigma_majorant
     majorant_inv = ONE / self % xsData % getMajorantXS(p)
 
+    ! Should never happen! Prevents NaN distances
+    if (majorant_inv /= majorant_inv) call fatalError(Here, "Majorant is 0")
+
     DTLoop:do
       distance = -log( p% pRNG % get() ) * majorant_inv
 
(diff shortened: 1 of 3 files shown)


Mikolaj-A-Kowalski (approved):
Thanks for the fix.

TransportOperator/transportOperatorDT_class.f90
```diff
@@ -49,6 +49,9 @@ subroutine deltaTracking(self, p, tally, thisCycle, nextCycle)
     ! Get majornat XS inverse: 1/Sigma_majorant
     majorant_inv = ONE / self % xsData % getMajorantXS(p)
 
+    ! Should never happen! Prevents NaN distances
+    if (majorant_inv /= majorant_inv) call fatalError(Here, "Majorant is 0")
+
     DTLoop:do
       distance = -log( p% pRNG % get() ) * majorant_inv
```
valeriaRaffuzzi, line 56:
```suggestion
    DTLoop:do
      distance = -log( p% pRNG % get() ) * majorant_inv

      ! Should never happen! Prevents NaN distances    
      if (distance /= distance) call fatalError(Here, "Majorant is 0")
```

Is this better @Mikolaj-A-Kowalski ?

Mikolaj-A-Kowalski: Or we can check if the majorant is `0` so its inverse is `Infinity`? I don't know...  
I have a feeling that here we have a small chance to get a NaN. Really somebody would have to have `0` majorant and also sample `0` from RNG to get `0 * Inf` expression.  
```suggestion
    ! Should never happen! Prevents Inf distances
    if (abs(majorant_inv) > huge(majorant_inv)) call fatalError(Here, "Majorant is 0")

    DTLoop:do
      distance = -log( p% pRNG % get() ) * majorant_inv

```

valeriaRaffuzzi: Yes sounds good!

TransportOperator/transportOperatorHT_class.f90
```diff
@@ -94,6 +94,8 @@ subroutine deltaTracking(self, p, tally, thisCycle, nextCycle)
     ! Get majornat XS inverse: 1/Sigma_majorant
     majorant_inv = ONE / self % xsData % getMajorantXS(p)
 
+   ! Should never happen! Prevents Inf distances
+    if (abs(majorant_inv) > huge(majorant_inv)) call fatalError(Here, "Majorant is 0")
     DTLoop:do
```
valeriaRaffuzzi, line 99:
```suggestion
    if (abs(majorant_inv) > huge(majorant_inv)) call fatalError(Here, "Majorant is 0")

    DTLoop:do
```

TransportOperator/transportOperatorHT_class.f90
```diff
@@ -148,6 +151,10 @@ subroutine surfaceTracking(self, p, tally, thisCycle, nextCycle)
       else
         sigmaT = self % xsData % getTransMatXS(p, p % matIdx())
         dist = -log( p % pRNG % get()) / sigmaT
+
+        ! Should never happen! Catches NaN distances
+        if (dist /= dist) call fatalError(Here, "Distance is NaN")
+
```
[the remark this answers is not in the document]
valeriaRaffuzzi: Yes, I can make an issue. Honestly it should be the same,  but I guess for the sake of not repeating too much code in the transport operators it makes sense.

PR #89, opened 2023-09-19
Update User Manual.rst

anuj-git-coding:
Lines 1084-1085

Added grid unstruct and grid predef to examples after discussion with bosses.


anuj-git-coding:
> Thanks a lot :D feel free to make a PR also for the other bugs that you said you found!

Turns out I was using an older version of the manual. The latest version does not contain that error. In the line below, "material" should be "mat". However, this is already fixed in the latest version.

cell1 { id 1; type simpleCell; surfaces (1 -6 90); filltype material; material fuel; }

Issue #92, opened 2023-09-20
Nuclide cache a bit broken


valeriaRaffuzzi:
A recent bug found in the NuclideCache logic: it could happen that nucCache % E_tot and nucCache % E_tail were different. When looking for microscopic xss, the code would occasionally check on nucCache % E_tail, which was correct, and as a consequence get the wrong total xs, that was updated earlier.

For now, checks on E_tot were added wherever E_tail is checked. In the future, we could separate the caches: one for the total xs and one for all the other xss, such to avoid confusion on what E_tot and E_tail refer to.

PR #94, opened 2023-09-20
Update check on energy values in nucCache

valeriaRaffuzzi:
Check total energy together with tail energy in the nuclide cache. 

Partially addresses issue #92.


Mikolaj-A-Kowalski (approved):
Thanks be to thee :bow: 
This was really a sneaky one :snake:

PR #95, opened 2023-09-20
Ecco33

valeriaRaffuzzi:
Adding ecco33 to predef energy grids.


Mikolaj-A-Kowalski (approved):
Looks good! 
I did not double check the bin boundaries though.

valeriaRaffuzzi:
> Looks good! I did not double check the bin boundaries though.

They come straight from Serpent, so they should be right!

PR #99, opened 2023-11-01
Bug in variance reduction

valeriaRaffuzzi:
In neutronCEimp, the variance reduction inside 'cutoffs' was going on even if the particle was dead. Incredibly I never spotted it, but it was definitely biasing and slowing down things a lot.


Mikolaj-A-Kowalski (approved):
Looks good. Thanks for the fix!

PR #100, opened 2023-11-18
Handle virtual collisions [fixed source and transportDT only]

martin220199:
It seems that both collisionclerk and transportoperatorDT have to know about virtual collisions wanting to be handled to 1) use majorant XS for scoring flux and 2) reporting both physical and virtual collisions (but avoid actually colliding when particle located in void). This is done by setting flag in collisionclerk based on input and communicating this throught tallyadmin to transportoperatorDT.

Virtual collisions can be handled by setting handleVirtual 1; in the input file:
tally  {
  collision_estimator { 
  type collisionClerk; response (flux); flux { type fluxResponse; }
  map {type energyMap; grid lin; min 0.0; max 14.1; N 70;}
  handleVirtual 1;
  }
}

Only values 1 or 0 allowed, 0 by default.

### DIFF
--- a/Tallies/TallyClerks/collisionClerk_class.f90
+++ b/Tallies/TallyClerks/collisionClerk_class.f90
@@ -59,6 +59,7 @@ module collisionClerk_class
 
     ! Useful data
     integer(shortInt)  :: width = 0
+    logical(defBool)   :: virtual = .false.
 
   contains
     ! Procedures used during build
@@ -114,6 +115,9 @@ contains
 
     ! Set width
     self % width = size(responseNames)
+
+    ! Handle virtual collisions
+    call dict % getOrDefault(self % virtual,'handleVirtual', .false.)
 
   end subroutine init
 
@@ -178,16 +182,25 @@ contains
   !!
   !! See tallyClerk_inter for details
   !!
-  subroutine reportInColl(self, p, xsData, mem)
+  subroutine reportInColl(self, p, xsData, mem, virtual)
     class(collisionClerk), intent(inout)  :: self
     class(particle), intent(in)           :: p
     class(nuclearDatabase), intent(inout) :: xsData
     type(scoreMemory), intent(inout)      :: mem
+    logical(defBool), intent(in)          :: virtual
     type(particleState)                   :: state
     integer(shortInt)                     :: binIdx, i
     integer(longInt)                      :: adrr
     real(defReal)                         :: scoreVal, flx
     character(100), parameter :: Here =' reportInColl (collisionClerk_class.f90)'
+
+    ! Calculate flux sample based on physical or virtual collision
+    if (self % virtual) then
+      flx = ONE / xsData % getMajorantXS(p)
+    else
+      if (virtual) return
+      flx = ONE / xsData % getTotalMatXS(p, p % matIdx())
+    end if
 
     ! Get current particle state
     state = p
@@ -209,9 +222,6 @@ contains
 
     ! Calculate bin address
     adrr = self % getMemAddress() + self % width * (binIdx -1)  - 1
-
-    ! Calculate flux sample 1/totXs
-    flx = ONE / xsData % getTotalMatXS(p, p % matIdx())
 
     ! Append all bins
     do i=1,self % width
--- a/Tallies/TallyClerks/tallyClerk_inter.f90
+++ b/Tallies/TallyClerks/tallyClerk_inter.f90
@@ -216,21 +216,23 @@ contains
   !!
   !! Process incoming collision report
   !!
-  !! See tallyAdmin_class for implicit assumptionas about the report.
+  !! See tallyAdmin_class for implicit assumptions about the report.
   !!
   !! Args:
-  !!   p [in]        -> Partice
-  !!   xsData [inout]-> Nuclear Database with XSs data
-  !!   mem [inout]   -> Score Memory to put results on
+  !!   p [in]         -> Partice
+  !!   xsData [inout] -> Nuclear Database with XSs data
+  !!   mem [inout]    -> Score Memory to put results on
+  !!   virtual [in]   -> Flag indicating virtual collision
   !!
   !! Errors:
   !!   Depend on specific Clerk
   !!
-  subroutine reportInColl(self,p, xsData, mem)
+  subroutine reportInColl(self,p, xsData, mem, virtual)
     class(tallyClerk), intent(inout)      :: self
     class(particle), intent(in)           :: p
     class(nuclearDatabase), intent(inout) :: xsData
     type(scoreMemory), intent(inout)      :: mem
+    logical(defBool), intent(in)          :: virtual
     character(100),parameter    :: Here = 'reportInColl (tallyClerk_inter.f90)'
 
     call fatalError(Here,'Report was send to an instance that does not support it.')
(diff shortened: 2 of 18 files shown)


ChasingNeutrons:
Hi Martin, thanks for doing this! It will be a great feature to have.

However, I think we can probably do it in a more generic way that hopefully means it will be agnostic to tally types (and number) and physics packages.

In particular, I think essentially only the transport operators, collisionClerk, and nuclearDatabases would need to be changed (or, following my second suggestion, the transport operators, collisionClerk, and particle). However, the former would require a slight redefinition of some cross sections and the functions to get and update them. 

One suggestion is we define a transport XS which would be set by either getMajorantXS or getTransMatXS. Both of these functions would set and return a XS value in the majorantCache (we might want to rename this to be the transportCache?). Naturally, the former would set this as the majorant, while the latter would set it as the total. Then, the collision clerk could call a different function which simply returns that XS without updating XSs any further (maybe returnTransXS, a new member of the nuclearDatabases - the interface, the ce database, and the baseMG database) to be used for scoring during reportInColl. This is more annoying for multigroup/baseMG, unfortunately, as it does not presently use a XS cache. I'm not sure what the optimal way around this is. Perhaps @valeriaRaffuzzi and @Mikolaj-A-Kowalski might be able to comment?

The changes this would require to transport operators would be making sure to call the appropriate get...XS in the right places, though these should mostly be fine already I believe. This would simply mean that DT would have getTotalXS calls when finding the total XS in the local material and not getTransMatXS calls there. Likewise, ST would have getTransMatXS calls during tracking (which I think it has already). 

Alternatively, we could simply add a flag to the particle, e.g., deltaFlight = .true./.false. . This would be used to decide whether to call getMajorant or getTotal during reportInColl and would be set as true or false during the start of a transit. I think this is the easiest option and probably the one to do for now at least! If we didn't do virtual scoring, this could be set to false at all times.

Finally, this means that whether to do virtual scoring could be an input only to the transportOperator. For now this can simply be a logical, rather than a number, but in future we could also do things like Jaakko does, i.e., set a maximum flight length before forcing a virtual collision.

Does this make sense?

martin220199:
Thank you for the detailed feedback! This makes sense and will, for now, look at the second option you proposed.

martin220199:
I squashed my 2 commits locally and pushed but seems maybe it was not updated here?

TransportOperator/transportOperator_inter.f90
```diff
@@ -123,8 +129,11 @@ end subroutine transport
   subroutine init(self, dict)
     class(transportOperator), intent(inout)  :: self
     class(dictionary), intent(in)            :: dict
+    logical(defBool)                         :: handleVirtual
 
-    ! Do nothing
+    ! Handle virtual collisions
+    call dict % getOrDefault(handleVirtual,'handleVirtual', .false.)
+    if (handleVirtual) self % handleVirtual = handleVirtual
```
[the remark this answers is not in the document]
valeriaRaffuzzi: Given that you added this property to the interface, you could modify the HT operator too (in the DT part).

TransportOperator/transportOperatorDT_class.f90
```diff
@@ -66,18 +66,24 @@ subroutine deltaTracking(self, p, tally, thisCycle, nextCycle)
...
+      end if
 
       ! Give error if the particle somehow ended in an undefined material
       if (p % matIdx() == UNDEF_MAT) then
         print *, p % rGlobal()
         call fatalError(Here, "Particle is in undefined material")
       end if
 
+      ! Accept both virtual and real collisions if virtualMode on
+      if (p % deltaFlight) exit DTLoop
```
valeriaRaffuzzi, line 81: I don't think this is what we're going for. If you exit the loop, you'll go to collisionOperator and process the collision as if it was always real. I think you should check if the collision is real or virtual, and if virtual you call reportInColl. If real you exit the loop and go to collisionOperator as normal.

valeriaRaffuzzi (changes requested):

valeriaRaffuzzi:
> I squashed my 2 commits locally and pushed but seems maybe it was not updated here?

I can see two commits, but maybe something got left behind?

TransportOperator/transportOperatorDT_class.f90
martin220199: Yes, my bad. Thanks for the feedback!

docs/User Manual.rst
```diff
@@ -218,6 +218,9 @@ The possible types are:
     energy, the ratio between the local material cross section and the majorant cross
     section is larger than the cutoff, DT is used; otherwise ST is used.
 
+* handleVirtual (*optional*, default = 0): if set to 1 handles virtual as well as physical
+  collisions
+
```
valeriaRaffuzzi, line 223: Now everything looks good, but I would fix the documentation a bit. I would move handleVirtual to the same level as '- cutoff', and add it under transportOperatorDT as well as transportOperatorHT. 
Also, in the description maybe it's more precise to say: 'if set to 1, delta tracking virtual collisions are tallied with a collisionClerk as well as physical collisions'

martin220199: Cheers, fixed now.

valeriaRaffuzzi (approved):
Now it looks good to me! Thanks a lot for implementing this, it will be very useful to have.

Tallies/TallyClerks/collisionClerk_class.f90
```diff
@@ -178,17 +182,26 @@ end function getSize
...
     integer(longInt)                      :: adrr
     real(defReal)                         :: scoreVal, flx
     character(100), parameter :: Here =' reportInColl (collisionClerk_class.f90)'
 
+    ! Calculate flux sample based on physical or virtual collision
+    if (self % virtual) then
+      flx = ONE / xsData % getMajorantXS(p)
+    else
+      if (virtual) return
+      if (.not. virtual) flx = ONE / xsData % getTotalMatXS(p, p % matIdx())
```
Mikolaj-A-Kowalski, line 202:
```suggestion
      flx = ONE / xsData % getTotalMatXS(p, p % matIdx())
```
Since we know that if virtual is true we `return` in the previous line I don't think this `if` is necessary.

Tallies/TallyClerks/tallyClerk_inter.f90
```diff
@@ -226,11 +226,12 @@ end subroutine print
   !! Errors:
   !!   Depend on specific Clerk
   !!
```
Mikolaj-A-Kowalski, line 228: I think we need to update the doc-comment of the interface as well?

Tallies/tallyAdmin_class.f90
```diff
@@ -442,16 +442,17 @@ end subroutine print
   !! Errors:
   !!   None
   !!
```
Mikolaj-A-Kowalski, line 444: Also doc-comment should probably be updated.

Mikolaj-A-Kowalski (commented):
Looks good! Thanks for the change just couple of minor, cosmetic suggestions. 
I feel really bad asking you for more... 
But maybe we should add a test case that involves virtual collisions and `handleVirtual` set to `.true.` in `collisionClerk_test`. If you don't have time to deal with it, I can do it. 
I just feel a bit uneasy without having tests to verify that the logic mapping setting to right flux estimate.

PR #101, opened 2023-11-20
More configuraton options in fissionSource

### LINKED ISSUE #96: Initialise problem
I got an error of initialization when running the involute reactor: 'Infinite loop in sampling of fission sites. Please check that defined volume contains fissile material.' The ( i > 200) might be too small. We can have a larger i, or we can make it un option for user to change.

Mikolaj-A-Kowalski:
Resolves #96. 
Add few options and makes cosmetic changes to the `fissionSource` 
* Changes the default number of re-sample limit from 200 to 10000
* Allows user to change the allowed number of re-sample attempts 
* Allows user to specify AABB where to try to place the points. 
* Add *User Manual* entry for the fission source that seems to be missing.

### DIFF
--- a/docs/User Manual.rst
+++ b/docs/User Manual.rst
@@ -203,9 +203,10 @@ Hence, an input would look like: ::
 fissionSource
 #############
 
-A source intended to be used in eigenvalue calculations. It is a type of volumetric source, which
-uniformly distributes fission sites in the geometry. The energy spectrum of the fission neutrons
-is based on a fixed incident energy provided by the user. The properties of a fission source are:
+A source intended to initialise eigenvalue calculations. If it is not defined in the input file, it is
+used with the default settings. It is a type of volumetric source, which uniformly distributes fission
+sites in the geometry. The energy spectrum of the fission neutrons is based on a fixed incident
+energy provided by the user. The properties of a fission source are:
 
 * data (*optional*, default='ce'): data type for source particles. Can be ``ce``
   or ``mg``.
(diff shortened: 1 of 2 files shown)


docs/User Manual.rst
```diff
@@ -185,21 +185,41 @@ It is a type of volumetric source. For the moment it is constrained to neutrons.
...
-      source { type materialSource; mat myMat; data ce; E 2.0; 
+      source { type materialSource; mat myMat; data ce; E 2.0;
       boundingBox (-5.0 -3.0 2.0 5.0 4.0 3.0); }
 
+fissionSource
+#############
+
+A source intended to be used in eigenvalue calculations. It is a type of volumetric source, which
+uniformly distributes fission sites in the geometry. The energy spectrum of the fission neutrons
+is based on a fixed incident energy provided by the user. The properties of a fission source are:
```
valeriaRaffuzzi, line 208: Maybe the first sentence is a bit misleading, it sounds like one should explicitly define it when doing an eigenvalue calculation. We could clarify a bit? This is just a proposal:

```suggestion
A source intended to initialise eigenvalue calculations. If it is not defined in the input file, it is 
used with the default settings. It is a type of volumetric source, which uniformly distributes fission
sites in the geometry. The energy spectrum of the fission neutrons is based on a fixed incident 
energy provided by the user. The properties of a fission source are:
```

Mikolaj-A-Kowalski: Sure! Commited (and removed trailing whitespace)

valeriaRaffuzzi (approved):
Nice, always good to have extra options :)

PR #103, opened 2023-11-22
User defined colours for plotting materials

### LINKED ISSUE #80: Material colourmaps
It is well know by this point that the Knuth hash maps index 1 (usually water or other coolant in the reactor) into a very unappealing shade of brown in 24-bit colour space. Perhaps to increase the visual appeal of the plots we should introduce a way of specifying a colormaps. 

The main question is if the user defined 24bit colours should be defined in the `materials` dictionary or in the separate dictionary in the bmp visualisation definition. Personally I am more inclined to the latter option.

Mikolaj-A-Kowalski:
Resolves #80. 
Makes the following changes: 
- Adds `offset` option to BMP output that allows to "shift" the colour sequence by some +ve or -ve integer. Thus change the colours in the image 
- Materials can now take optional `rgb` entry which specifies the colour which will be used to print them. e.g. `rgb (255 0 0)`. 
- Minor stylistic and grammatical changes  
- Update to the user manual and `c5g7_mox` example input file to showcase new features

### DIFF
--- a/NuclearData/materialMenu_mod.f90
+++ b/NuclearData/materialMenu_mod.f90
@@ -8,6 +8,7 @@
 !! Public Members:
 !!   materialDefs -> array of material definitions of type materialItem
 !!   nameMap      -> Map that maps material name to matIdx
+!!   colourMap    -> Map that maps matIdx to 24bit colour (to use for visualisation)
 !!
 !! Interface:
 !!   init      -> Load material definitions from a dictionary
@@ -21,8 +22,10 @@
 module materialMenu_mod
 
   use numPrecision
-  use universalVariables, only : NOT_FOUND, VOID_MAT, OUTSIDE_MAT
+  use universalVariables, only : NOT_FOUND, VOID_MAT, OUTSIDE_MAT, UNDEF_MAT
   use genericProcedures,  only : fatalError, charToInt, numToChar
+  use colours_func,       only : rgb24bit
+  use intMap_class,       only : intMap
   use charMap_class,      only : charMap
   use dictionary_class,   only : dictionary
 
@@ -83,6 +86,7 @@ module materialMenu_mod
   !!       5010.03  2.0E-005;
   !!     }
   !!     xsFile /home/uberMoffTarkin/XS/mat1.xs;
+  !!     #rgb (255 0 0); # // RGB colour to be used in visualisation
   !!   }
   !!
   !! NOTE: the moder dictionary is optional, necessary only if S(a,b) thermal scattering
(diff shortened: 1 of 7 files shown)


NuclearData/materialMenu_mod.f90
```diff
@@ -300,6 +319,17 @@ subroutine init_materialItem(self, name, dict)
...
     end do
 
+    ! Add colour info if present
+    if(dict % isPresent('rgb')) then
+      call dict % get(temp, 'rgb')
+
+      if (size(temp) /= 3) then
+        call fatalError(Here, "'rgb' keyword must have 3 values")
+      end if
+      print *, "HERE", temp, self % matIdx
```
valeriaRaffuzzi, line 329: This looks like a debugging remanence?

docs/User Manual.rst
```diff
@@ -768,6 +774,7 @@ Example 1: ::
       8016.03    0.018535464; }
       }
       water { temp 273;
+      rgb (0 0 200); // Blue colour is used for water in plots
```
valeriaRaffuzzi, line 777: I'm actually not a big fan of the comments in the manual, the text should be enough to explain everything.

valeriaRaffuzzi (commented):
If the offset is there to get rid of the brown, could we just sample it randomly per each material? It's sort of difficult for a user to know what's a good value for the offset anyway.

docs/User Manual.rst
Mikolaj-A-Kowalski: Ok. I will remove it then ;-) 
I think you are right. It is needless repetition of the data in the paragraph, which has a good chance to go out of date.

NuclearData/materialMenu_mod.f90
Mikolaj-A-Kowalski: It very much is! Good catch!

Mikolaj-A-Kowalski:
> If the offset is there to get rid of the brown, could we just sample it randomly per each material? It's sort of difficult for a user to know what's a good value for the offset anyway.

Personally I really don't like the colours being assigned at random for each program execution. I though the `offset` is a good compromise there since it is not too difficult to key in few random digits and if one finds the colour scheme that looks nice, it would be easy to re-use it. At least this was my thinking. 

That being said assigning the offset from some source of randomness (probably hashing the time like it is done for RNG) could be added without too much difficulty if it is a desirable feature. Part of my reluctance may be just trying to be 'not like Serpent' ;-)

valeriaRaffuzzi (changes requested):
See comments.

Mikolaj-A-Kowalski:
I believe all outstanding issues should have been addressed now. 
Also now that offset is not specified, it is determined by hashing current date & time like for the seed in a calculation.

Visualisation/visualiser_class.f90
```diff
@@ -1,394 +1,398 @@
...
+
+    ! Get name of the output file
+    call dict % get(outputFile, 'output')
+    outputFile = trim(outputFile) // '.bmp'
+
+    ! Central point
+    call dict % get(temp, 'centre')
+
+    if (size(temp) /= 3) then
+      call fatalError(Here, "'center' must have size 3. Has: "//numToChar(size(temp)))
```
valeriaRaffuzzi, line 253:
```suggestion
      call fatalError(Here, "'centre' must have size 3. Has: "//numToChar(size(temp)))
```

Let's not confuse possible oversea users!!

valeriaRaffuzzi (approved):
Looks great! Can't wait to make a non brown plot!!!

There's only that one typo in the visualiser but otherwise good to go

PR #104, opened 2023-12-05
Optimisation of MG nuclear data 

valeriaRaffuzzi:
A material cache for multi group data was added, similar to the CE one. The cache is read or updated when getting the total material xss from the baseMgNeutronDatabase or when getting all the xss from baseMgNeutronMaterial_inter. 

Additionally, a material pointer is also stored in the cache; this is initialised when activating the database. 

On a 2D PWR pincell problem run on fermiac on 120 cores, perf reported ~50% time spent and ~50% cache misses happening when retrieving the total xs, and ~10% time and cache misses when getting the material pointer from the database.

After adding the cross section and pointer cache, those overheads and cache misses disappeared and the code became ~60% faster.

### DIFF
--- a/NuclearData/mgNeutronData/mgNeutronCache_mod.f90
+++ b/NuclearData/mgNeutronData/mgNeutronCache_mod.f90
@@ -1,0 +1,92 @@
+!!
+!! Global Cache for MG Neutron Nuclear Data
+!!
+!! Any active nuclear database writes XSs to this module.
+!! Any material handle reads from this module
+!!
+!! Public Members:
+!!   matCache      -> Array of cached data for materials
+!!   majorantCache -> Array of cached data for majorant XS
+!!
+!! NOTE:
+!!   Cache arrays are deliberatly not targets. This is becouse there should be no pointers to the
+!!   cache. Any update call can change energy of any value so it would not be possible that the
+!!   energy of XSs pointed by pointers would not change silently.
+!!
+module mgNeutronCache_mod
+
+  use numPrecision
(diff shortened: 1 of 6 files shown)


NuclearData/mgNeutronData/mgNeutronCache_mod.f90
```diff
@@ -0,0 +1,92 @@
...
+
+  !!
+  !! Structure that contains cached data for each MG Neutron Material
+  !!
+  !! Public Members:
+  !!   G_tot  -> Energy group of the total XS in xss
+  !!   G_tail -> Energy group of all XSs in xss except total
+  !!   xss    -> Cached Cross-Section values
+  !!
+  type, public :: cacheMatDat
```
Mikolaj-A-Kowalski, line 34: Add pointer member to the `Public Members` section? 
We can also say in the doc comment that it is included here because it makes scalability better, perhaps? I imagine that perhaps someone might be confused why it is stored here and not retrieved via the `getMaterial` procedure?

Mikolaj-A-Kowalski (approved):
Looks great! Thanks for the change! I think it can be merged without a problem. 
Out of curiosity I will try to run some test cases to figure why why it works exactly, but hopefully one scalability issue is down!

valeriaRaffuzzi:
> Looks great! Thanks for the change! I think it can be merged without a problem. Out of curiosity I will try to run some test cases to figure why why it works exactly, but hopefully one scalability issue is down!

Yes, it's definitely worth trying to understand what's going on! The docs should be updated now, if you're ok with it feel free to merge :)

PR #106, opened 2024-01-17
DT majorant

valeriaRaffuzzi:
This PR adds the precalculation of the majorant xss on a unionised grid for delta tracking. 

Few things to notice: 
- I decided to have the transportOperator call the initialisation procedure in the database, so that only DT and HT call it and ST doesn't
- Precalculated majorant is the new default and only option available when using DT
- This way of constructing the grid is the easiest, but it's probably not the fastest. For instance, constructing the unionised energy grid (which is not really a parallelisable thing) for the CE c5g7 takes ~8 seconds (while calculating the majorant on that grid takes a fraction of a second). But then transport is x2 times faster with DT. If you have a better way in mind let me know! For now I wanted this done fast. 
- This was tested and it works fine, and the relevant integrationTests were adapted too to include the change.


NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
```diff
@@ -328,22 +336,26 @@ subroutine updateMajorantXS(self, E, rand)
...
+      if(idx <= 0) then
+        call fatalError(Here,'Failed to find energy: '//numToChar(E)//&
+                             ' in unionised majorant grid')
+      end if
+
+      associate(E_top => self % eGrid(idx + 1), E_low  => self % eGrid(idx))
+        f = (E - E_low) / (E_top - E_low)
+      end associate
+
+      maj % xs = self % majorant(idx+1) * f + (ONE - f) * self % majorant(idx)
```
Mikolaj-A-Kowalski, line 357: I have a fear that if it happens that the `initMajorant` is not called by the user, but for some reason a call to `updateMajorantXS` is made, then the code will segment on the binary search. 

Similarly in the case a following calls are made: 
- activate with  materials 1,2,3 
- init majorant 
- 2nd activate (reactivate) with materials 1,2,3,4 
Then it is possible that the majorant will find itself in an invalid state (not representing the expected majorant).  
What we do with this will depend on what we finally decide on having `initMajorant` as a separate procedure in the interface, but it may involve some `allocated(self % eGrid)` checks to give more descriptive error message)

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
```diff
@@ -771,13 +783,122 @@ subroutine activate(self, activeMat)
...
+          call quickSort(tmpGrid)
+          tmpGrid = removeDuplicatesSorted(tmpGrid)
+
+          ! Add nuclide to the set
+          call nucMap % add(nucIdx, IN_SET)
+
+        end if
+
+      end do
+    end do
```
Mikolaj-A-Kowalski, line 855: It is perhaps not at this moment but the (relatively) low hanging fruits for optimisation here would be to create a `real` version of the `dynArray_class` and use it as the `tmpGrid`. I guess it could take care of a large portion of the runtime by avoiding a lot of reallocation. 

Of course `dynArray` would than also have in-place versions of `removeDuplicates` and `quickSort` added.

NuclearData/mgNeutronData/baseMgNeutron/baseMgNeutronDatabase_class.f90
```diff
@@ -330,15 +332,50 @@ end subroutine init
...
-    self % activeMats = activeMat
+    ! Set build console output flag
+    if(present(silent)) then
+      loud = .not.silent
+    else
+      loud = .true.
+    end if
 
-    ! Precalculate majorant xs for delta tracking
+    if (loud) print '(A)', 'Building unionised majorant cross section'
```
Mikolaj-A-Kowalski, line 373:
```suggestion
    if (loud) print '(A)', 'Building MG unionised majorant cross section'
```
Maybe we should highlight that it is a MG case?

TransportOperator/transportOperatorHT_class.f90
```diff
@@ -81,7 +70,9 @@ subroutine tracking_selection(self, p, tally, thisCycle, nextCycle)
 
   end subroutine tracking_selection
 
-
+  !!
+  !! Performs delta tracking until a real collision point is found
+  !!
```
Mikolaj-A-Kowalski, line 75: Long necessary and long forgotten part! :-) 
Thanks for filling the gap!

TransportOperator/transportOperatorHT_class.f90
```diff
@@ -138,7 +129,9 @@ subroutine deltaTracking(self, p, tally, thisCycle, nextCycle)
     call tally % reportTrans(p)
   end subroutine deltaTracking
 
-
+  !!
+  !! Performs surface tracking until a collision point is found
+  !!
```
Mikolaj-A-Kowalski, line 134: As above! ;-)

Mikolaj-A-Kowalski (changes requested):
Thanks for the changes! They look great! I suppose that the optimisation of the unionised grid construction we wish to add for later (and open an issue about it)? Unless do you prefer to try to trim some fat in this PR? 

The only major point is what we discussed earlier :-) My instinct would be to have a unionisation as a option in the nuclearDatabase configuration dictionary. Then `initMajorant` would be called (or not) in the `activate` procedure. Also the `updateMajorantXS` would decide between old and the unionised grid search based on the allocation status of the `self % majorant`. 

Part of my fear is that with `initMajorant` and `activate` being separate functions weird things may be happening (like the example I outlined in the code attached comment) in case of some sequences of calls and we would need to add checks to protect against those situations (which is not difficult, but there is always a danger of missing some nasty, specific call sequence). 

Also I might have missed something but does the unionised grid for majorant require that URR is switched off for all nuclides?

ChasingNeutrons:
I agree with Mikolaj's comments. Related to that, I think, until the majorant construction is optimised, it is desirable to be able to switch off precomputing a majorant in the input file - otherwise we might encounter a nasty case with many nuclides that takes impractically long to run just from constructing the grid which would run just fine at the moment. It might require switching off majorant construction and recompiling.

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
valeriaRaffuzzi: It is a bit involved, so I will just open an issue. It will be needed for actual use with burnup or so.

valeriaRaffuzzi:
Issue #107 is addressed here. The unionised energy grid is build gradually from the nuclides grids instead of concatenating arrays and using quicksort. Now it takes a fraction of a second to precompute everything instead of ~10s.

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
```diff
@@ -3,11 +3,13 @@ module aceNeutronDatabase_class
...
   use endfConstants
   use universalVariables
-  use genericProcedures, only : fatalError, numToChar
-  use dictionary_class,  only : dictionary
-  use RNG_class,         only : RNG
-  use charMap_class,     only : charMap
-  use intMap_class,      only : intMap
+  use errors_mod,         only : fatalError
+  use genericProcedures,  only : numToChar, concatenate, quickSort, &
+                                 removeDuplicatesSorted, binarySearch
```
valeriaRaffuzzi, line 8:
```suggestion
  use genericProcedures,  only : numToChar, removeDuplicatesSorted, binarySearch
```

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
```diff
@@ -51,19 +52,26 @@ module aceNeutronDatabase_class
   !! It's possible to use probability tables in the unresolved resonance range if
   !! ures is included in the input file
   !!
+  !! NOTE: the unionised majorant is not calculated and used if probability tables
```
ChasingNeutrons, line 55: I think we should open an issue on this. It will be very desirable to still have a unionised majorant, e.g., if we want to accurately run fast problems. I understand it shouldn't have a big impact for thermal problems, even with burnup.

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
```diff
@@ -51,19 +52,26 @@ module aceNeutronDatabase_class
...
   !!   nuclides   -> array of aceNeutronNuclides with data
   !!   materials  -> array of ceNeutronMaterials with data
   !!   Ebounds    -> array with bottom (1) and top (2) energy bound
+  !!   majorant   -> unionised majorant cross section
+  !!   eGridUnion -> unionised energy grid
   !!   activeMat  -> array of materials present in the geometry
   !!   nucToZaid  -> map to link nuclide index to zaid index
   !!   hasUrr     -> ures probability tables flag, it's false by default
   !!   hasDBRC    -> DBRC flag, it's false by default
+  !!   has majorant -> unionised majorant cross section flag
```
ChasingNeutrons, line 74: hasMajorant, not has majorant

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
```diff
@@ -653,6 +692,9 @@ subroutine init(self, dict, ptr, silent )
       self % Ebounds(2) = min(self % Ebounds(2), self % nuclides(i) % eGrid(j))
     end do
 
+    ! Read unionised majorant flag
+    call dict % getOrDefault(self % hasMajorant, 'majorant', .false.)
```
ChasingNeutrons, line 696: I would be inclined to set this to true by default. We run DT often and it's one less flag to set in that case, because we would always want the precomputed majorant (as in MG). It's also a very small price to pay if we happen to be running ST.

NuclearData/nuclearDatabase_inter.f90
```diff
@@ -196,7 +197,7 @@ end function getMaterial
     !! Allows to retrive an access to nuclide data for all databases types
```
ChasingNeutrons, line 197: While we're here.... retrieve*

TransportOperator/transportOperatorHT_class.f90
```diff
@@ -98,7 +89,7 @@ subroutine deltaTracking(self, p, tally, thisCycle, nextCycle)
     if (abs(majorant_inv) > huge(majorant_inv)) call fatalError(Here, "Majorant is 0")
 
     DTLoop:do
-      distance = -log( p% pRNG % get() ) * majorant_inv
+      distance = -log( p % pRNG % get() ) * majorant_inv
 
       ! Move partice in the geometry
```
ChasingNeutrons, line 94: particle*

docs/User Manual.rst
```diff
@@ -742,7 +742,9 @@ from ACE files.
   resonance probability tables treatment
 * DBRC (*optional*, default = no DBRC): list of ZAIDs of nuclides for which DBRC has
   to be applied.
-
+* majorant (*optional*, default = 0): 1 for true; 0 for false; flag to activate the
```
ChasingNeutrons, line 745: As above, I think this should default to true. I suppose it comes down to which would we prefer to suffer in a worst case - a slower calculation or more memory. I think the memory is more tolerable.

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
valeriaRaffuzzi: Sure, I'll open an issue. I am not sure if one can a priori find out what's the majorant when using probability tables. One approach could be to sample, say, 1000 times per each energy when calculating the majorant, and retaining the highest xs.

valeriaRaffuzzi:
Yet more changes to this PR: 

- now the majorant can be precalculated also with ures tables (solves issue  #106): the highest entry is the ures tables at each energy is also precomputed and used to calculated the overall majorant

- the majorant is nudged up by 1e-6 to avoid small floating point errors or inconsistencies that seemed to appear (e.g., majorant being smaller than the largest material xs by 1e-15)

- I added to the unionised grid the energy points corresponding to the limit energies of URR and S(a,b) tables. This is because I was getting a majorant < mat_xs just below the S(a,b) upper energy boundary: two consecutive energies were one within and one outside the S(a,b) range, so interpolating gave the wrong result. Honestly nothing breaks if this happens, because it's not very frequent. Stuff like these are hard to avoid when there are discontinuities in the xss like due to Sab or ures, and they could still happen. However now for the cases I tried it behaves ok.

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
```diff
@@ -754,30 +801,305 @@ subroutine init_DBRC(self, nucDBRC, nucSet, map)
...
+        ! Loop over nuclides
+        do k = 1, size(self % materials(matIdx) % nuclides)
+          dens     = self % materials(matIdx) % dens(k)
+          nucIdx   = self % materials(matIdx) % nuclides(k)
+
+          associate (nuc => self % nuclides(nucIdx))
+
+            needsUrr = (nuc % hasProbTab .and. E >= nuc % urrE(1) .and. E <= nuc % urrE(2))
+
+            ! Check if present nuclide uses URR tabes
```
ChasingNeutrons, line 1062: tables*

NuclearData/ceNeutronData/aceDatabase/aceNeutronNuclide_class.f90
```diff
@@ -741,14 +742,14 @@ subroutine init(self, ACE, nucIdx, database)
     ! Create a stack of MT reactions, devide them into ones that produce 2nd-ary
```
ChasingNeutrons, line 742: divide*

NuclearData/ceNeutronData/aceDatabase/aceNeutronNuclide_class.f90
```diff
@@ -741,14 +742,14 @@ subroutine init(self, ACE, nucIdx, database)
...
     ! particlues and pure absorbtion
```
ChasingNeutrons, line 743: particles*, absorption*

NuclearData/nuclearDatabase_inter.f90
```diff
@@ -217,15 +219,15 @@ end function getNuclide
     !!
     !! Return a pointer to a reaction
     !!
-    !! Allows to retrive an access to reaction data for all databases types
+    !! Allows to retrieve an access to reaction data for all databases types
     !! Reactions can be associated either with nuclides or materials. Thus, there is ambiguity
     !! whether material or nuclide should be asked to provide reaction data (using matIdx or nuIdx)
```
ChasingNeutrons, line 224: nucIdx*

ChasingNeutrons (approved):
I feel bad that my only comments are tiny typos... But this looks great! We're a good bit more respectable after this!

valeriaRaffuzzi:
> I feel bad that my only comments are tiny typos... But this looks great! We're a good bit more respectable after this!

Well, this is the result of several iterations already! Luckily we converged :)

Mikolaj-A-Kowalski (approved):
Looks good! Thanks for all the hard work. The URR added some complication to this one ;-)

PR #109, opened 2024-01-24
Sample with replacment for particles to copy in the dungeon

Mikolaj-A-Kowalski:
Until now, in case where the population was too little and some of the particles had to be copied, we sampled duplicates with replacement. This changes the algorithm to sample without replacement in that case as well. 

A reservoir sampling is used to select sample of particle indices that will be duplicated at the end of the list.


valeriaRaffuzzi (approved):
Thank you for this great contribution, Mikolaj! 

It looks much more rigorous than before.

PR #111, opened 2024-01-25
Fix related to previous varRed PR

valeriaRaffuzzi:
After removing the property splitCount from the particleState, the behavior when detaining a particleState was undefined. 

This way, every time a particle state is detained, the counter is reinitialised.


Mikolaj-A-Kowalski (approved):
Ups... my fault!

Issue #52, opened 2023-07-07
Multi-thread OpenMP reproducibility


Mikolaj-A-Kowalski:
At the moment SCONE gives reproducible results (same for the same seed) as long as it is run with a single thread. 
With multiple threads however reproducibility is not obtained as a result of the race condition when adding new particles to the dungeon for the next cycles (basically the order and hence index of particles is not deterministic). 

This is a well known problem in MC with OpenMP and can be solved by making sure the order of particles does not depend on the number of threads.

Note: 
The method is described in https://www.osti.gov/biblio/1417155 
Pages 341 - 348

valeriaRaffuzzi:
Addressed by #113

PR #113, opened 2024-01-27
OpenMP reproducibility

Mikolaj-A-Kowalski:
First of all I think this one still requires a bit of work and more testing. I verified that things work on my laptop, but I haven't properly studied the impact on scaling. 

Basically this PR makes the parallel calculations with OpenMP reproducible. The following main changes were required: 
- `source_inter` sampling procedure had to be modified to skip from the initial RNG state (not the current thread-private). I'm afraid this bug avoided death in PR before...
- The incorrect, thread-private, RNG was used for sampling of the renormalisation of the source 
- The `particleDungeon` sorting has been implemented based on the `showerID` which is basically the sequence in this cycle dungeon of the particle. I am not happy with the name but I couldn't figure out a better one! (Also addresses #52) 
  - The sorting is done with some extra memory allocated at each cycle to store the permutation and count arrays. 
  - The permutation is done in place to conserve memory requirements (but it makes parallelisation not possible :-/, I am not sure it was a correct decision) 
  - Generally the sorting and renormalisation of the `particleDungeon` are done in serial and I fear a bit about scalability as a result of it.    
  
  P.S. Also I realised now that I forgot to add extra tests :-/ Treat it like a draft for now, please.

### DIFF
--- a/ParticleObjects/Tests/particleDungeon_test.f90
+++ b/ParticleObjects/Tests/particleDungeon_test.f90
@@ -217,7 +217,7 @@ contains
     ! Store some particles with non-uniform weight
     do i = 1,10
       p % w = 0.5_defReal + i * 0.1_defReal
-      p % showerID = 1       ! Avoid triggering error on sort by showerID
+      p % broodID = 1       ! Avoid triggering error on sort by broodID
       call dungeon % detain(p)
     end do
 
@@ -255,7 +255,7 @@ contains
     ! Store some particles with non-uniform weight
     do i = 1,10
       p % w = 0.5_defReal + i * 0.1_defReal
-      p % showerID = 1       ! Avoid triggering error on sort by showerID
+      p % broodID = 1       ! Avoid triggering error on sort by broodID
       call dungeon % detain(p)
     end do
 
@@ -277,7 +277,7 @@ contains
   !! Test sorting of the population by shower ID without duplicates
   !!
 @Test
-  subroutine testSortingByShowerID()
+  subroutine testSortingByBroodID()
     type(particleDungeon)    :: dungeon
     type(particle)           :: p
     integer(shortInt)        :: i
@@ -288,27 +288,27 @@ contains
 
     ! Store some particles with shower ID in reverse order
     do i = 1,10
-      p % showerID = 10 - i + 1
+      p % broodID = 10 - i + 1
       call dungeon % detain(p)
     end do
 
     ! Sort by shower ID
-    call dungeon % sortByShowerId(10)
+    call dungeon % sortByBroodID(10)
 
     ! Verify order
     do i = 1,10
       call dungeon % copy(p, i)
-      @assertEqual(i, p % showerID)
+      @assertEqual(i, p % broodID)
     end do
 
-  end subroutine testSortingByShowerID
+  end subroutine testSortingByBroodID
 
 
   !!
   !! Test sorting of the population by shower ID with duplicates
   !!
   @Test
-  subroutine testSortingByShowerID_withDuplicates()
+  subroutine testSortingByBroodID_withDuplicates()
     type(particleDungeon)    :: dungeon
     type(particle)           :: p
     integer(shortInt)        :: i, j
@@ -323,25 +323,25 @@ contains
     ! that the insertion order is preserved (for particles with the same shower ID)
     do j = 1,N_duplicates
       do i = 1,10
-        p % showerID = 10 - i + 1
+        p % broodID = 10 - i + 1
         p % G = j
         call dungeon % detain(p)
       end do
     end do
 
     ! Sort by shower ID
-    call dungeon % sortByShowerId(10)
+    call dungeon % sortByBroodID(10)
 
     ! Verify order
     do i = 1,10
       do j = 1, N_duplicates
         call dungeon % copy(p, j + (i-1) * N_duplicates)
-        @assertEqual(i, p % showerID)
+        @assertEqual(i, p % broodID)
         @assertEqual(j, p % G)
       end do
     end do
 
-  end subroutine testSortingByShowerID_withDuplicates
+  end subroutine testSortingByBroodID_withDuplicates
 
 
 end module particleDungeon_test
--- a/ParticleObjects/Tests/particleDungeon_test.f90
+++ b/ParticleObjects/Tests/particleDungeon_test.f90
@@ -217,6 +217,7 @@ contains
     ! Store some particles with non-uniform weight
     do i = 1,10
       p % w = 0.5_defReal + i * 0.1_defReal
+      p % showerID = 1       ! Avoid triggering error on sort by showerID
       call dungeon % detain(p)
     end do
 
@@ -254,6 +255,7 @@ contains
     ! Store some particles with non-uniform weight
     do i = 1,10
       p % w = 0.5_defReal + i * 0.1_defReal
+      p % showerID = 1       ! Avoid triggering error on sort by showerID
       call dungeon % detain(p)
     end do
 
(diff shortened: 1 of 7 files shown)


ChasingNeutrons:
I must say showerID is quite a baffling name to me. What's the logic for choosing this name? This is just a unique ID for a given particle? Or if not, could you explain what it is supposed to do? Maybe something like queueID? Or threadID?

Mikolaj-A-Kowalski:
> I must say showerID is quite a baffling name to me. What's the logic for choosing this name? This is just a unique ID for a given particle? Or if not, could you explain what it is supposed to do? Maybe something like queueID? Or threadID?

So the name comes from my photon warped brain and the 'EM shower' term. Basically to denote a source particle and all its progeny (with fission assumed to be absorption).  It is not a uniqueID of a particle per se as it is can be shared by multiple particles and their values are reset each cycle. `queueID` may be better but still makes me think of a identity of some queue, not the correlated bunch of particles. 

Basically name is like it is because I couldn't think of anything good and 'shower' was the only thing I could think of and I didn't want to stall the development by not being able to think of a name. But I am very open for suggestions. I did presume that `showerID` will be a placeholder really.

ChasingNeutrons:
I feared it was something perverse like this! geneticID? familyID? spawnID (didn't that feature somewhere else?)?

ChasingNeutrons:
Also parentID? Maybe the most important thing is to have a comment in particle explaining what it is.

Mikolaj-A-Kowalski:
> I feared it was something perverse like this! geneticID? familyID? spawnID (didn't that feature somewhere else?)?

Perverse!!! 🤕... :cry: Auch! 
The problem is that I am not happy with any of them it is more like: `sequence of the parent particle in the Dungeon/source` but that does not exactly make a good variable name...

ChasingNeutrons:
It's certainly hard to make that compact. If it's just to denote a group of particle together (and we don't want to use the word 'group'!) then what about cohortID?

Mikolaj-A-Kowalski:
> It's certainly hard to make that compact. If it's just to denote a group of particle together (and we don't want to use the word 'group'!) then what about cohortID?

I am happy with the `cohortID`, I don't have better ideas to be honest. Unless @valeriaRaffuzzi has something in mind? 
Otherwise I will change `showerID` to `cohortID`

valeriaRaffuzzi:
It is like a familyID but just for one generation, right? broodID? 

Google tells me it's a collective name for a bunch of siblings.

ParticleObjects/particleDungeon_class.f90
```diff
@@ -541,10 +610,13 @@ subroutine printToFile(self, name)
 
     ! Print out each particle co-ordinate
     do i = 1, self % pop
-      write(10,'(8A)') numToChar(self % prisoners(i) % r), &
-                       numToChar(self % prisoners(i) % dir), &
-                       numToChar(self % prisoners(i) % E), &
-                       numToChar(self % prisoners(i) % G)
+      ! write(10,'(8A)') numToChar(self % prisoners(i) % r), &
```
valeriaRaffuzzi, line 613: Should we remove the commented bit here?

ParticleObjects/particle_class.f90
```diff
@@ -649,31 +654,6 @@ function equal_particleState(LHS,RHS) result(isEqual)
     end if
   end function equal_particleState
 
-!  !!
```
valeriaRaffuzzi, line 652: Love a good clean up!

PhysicsPackages/eigenPhysicsPackage_class.f90
```diff
@@ -195,6 +195,8 @@ subroutine cycles(self, tally, tallyAtch, N_cycles)
         ! Obtain particle current cycle dungeon
         call self % thisCycle % copy(neutron, n)
 
+        neutron % showerID = n
```
valeriaRaffuzzi, line 198: Could this line be included into `particleDungeon % copy` instead of here? I think it would look neater.

PhysicsPackages/fixedSourcePhysicsPackage_class.f90
```diff
@@ -181,6 +181,7 @@ subroutine cycles(self, tally, N_cycles)
 
         ! Obtain particle from dungeon
         call self % thisCycle % copy(p, n)
+        p % showerID = n
```
valeriaRaffuzzi, line 184: Same comment as before about the `copy`.

Mikolaj-A-Kowalski: Applied

PhysicsPackages/eigenPhysicsPackage_class.f90
Mikolaj-A-Kowalski: Applied

ParticleObjects/particleDungeon_class.f90
Mikolaj-A-Kowalski: Has been removed now

Mikolaj-A-Kowalski:
I think that all the comments have been addressed. 
Unrelated to this PR I added a commit to fix an issue with reuse of random numbers after source generation. it is not clean to put it here but was afraid I would forget otherwise ;-) 

I have also updated the documentation to highlight that the common buffer in fixed source calculation will (most likely) destroy OpenMP reproducibility.

Mikolaj-A-Kowalski:
> I think that all the comments have been addressed. Unrelated to this PR I added a commit to fix an issue with reuse of random numbers after source generation. it is not clean to put it here but was afraid I would forget otherwise ;-)

I think that when I did this change i did not think sufficiently about fixed source calculations. I think there would be a reuse then still.

Mikolaj-A-Kowalski:
> > I think that all the comments have been addressed. Unrelated to this PR I added a commit to fix an issue with reuse of random numbers after source generation. it is not clean to put it here but was afraid I would forget otherwise ;-)
> 
> I think that when I did this change i did not think sufficiently about fixed source calculations. I think there would be a reuse then still.

The commit regarding the reuse of random numbers was dropped.

valeriaRaffuzzi (approved):
It all looks good to me now! Thanks for working on this :)

ChasingNeutrons (approved):
One typo and then I can merge it in!

ParticleObjects/particle_class.f90
```diff
@@ -111,6 +103,7 @@ module particle_class
     integer(shortInt)          :: fate = 0       ! Neutron's fate after being subjected to an operator
     integer(shortInt)          :: type           ! Particle type
     integer(shortInt)          :: collisionN = 0 ! Index of the number of collisions the particle went through
+    integer(shortInt)          :: broodID = 0    ! ID of the shower (source particle number)
```
valeriaRaffuzzi, line 106:
```suggestion
    integer(shortInt)          :: broodID = 0    ! ID of the brood (source particle number)
```

ParticleObjects/Tests/particleDungeon_test.f90
```diff
@@ -271,5 +273,75 @@ subroutine testNormPopUp()
 
   end subroutine testNormPopUp
 
+  !!
+  !! Test sorting of the population by shower ID without duplicates
```
valeriaRaffuzzi, line 277:
```suggestion
  !! Test sorting of the population by brood ID without duplicates
```

ParticleObjects/Tests/particleDungeon_test.f90
```diff
@@ -271,5 +273,75 @@ subroutine testNormPopUp()
...
+    ! Initialise
+    call dungeon % init(10)
+
+    ! Store some particles with shower ID in reverse order
+    do i = 1,10
+      p % broodID = 10 - i + 1
+      call dungeon % detain(p)
+    end do
+
+    ! Sort by shower ID
```
valeriaRaffuzzi, line 295:
```suggestion
    ! Store some particles with brood ID in reverse order
    do i = 1,10
      p % broodID = 10 - i + 1
      call dungeon % detain(p)
    end do

    ! Sort by brood ID
```

ParticleObjects/Tests/particleDungeon_test.f90
```diff
@@ -271,5 +273,75 @@ subroutine testNormPopUp()
...
+    ! that the insertion order is preserved (for particles with the same shower ID)
+    do j = 1,N_duplicates
+      do i = 1,10
+        p % broodID = 10 - i + 1
+        p % G = j
+        call dungeon % detain(p)
+      end do
+    end do
+
+    ! Sort by shower ID
```
valeriaRaffuzzi, line 332:
```suggestion
  !! Test sorting of the population by brood ID with duplicates
  !!
  @Test
  subroutine testSortingByBroodID_withDuplicates()
    type(particleDungeon)    :: dungeon
    type(particleState)      :: p
    integer(shortInt)        :: i, j
    integer(shortInt), parameter :: N_duplicates = 7
    real(defReal), parameter :: TOL = 1.0E-9

    ! Initialise
    call dungeon % init(10 * N_duplicates)

    ! Store some particles with brood ID in reverse order
    ! Use the group number to distinguish duplicates and make sure
    ! that the insertion order is preserved (for particles with the same brood ID)
    do j = 1, N_duplicates
      do i = 1, 10
        p % broodID = 10 - i + 1
        p % G = j
        call dungeon % detain(p)
      end do
    end do

    ! Sort by brood ID
```

ParticleObjects/particleDungeon_class.f90
```diff
@@ -413,6 +424,10 @@ subroutine normSize(self, N, rand)
       call fatalError(Here,'Requested size: '//numToChar(N) //' is not +ve')
     end if
 
+    ! Determine the maximum shower ID and sort the dungeon
```
valeriaRaffuzzi, line 427:
```suggestion
    ! Determine the maximum brood ID and sort the dungeon
```

ParticleObjects/particleDungeon_class.f90
```diff
@@ -456,6 +471,69 @@ subroutine normSize(self, N, rand)
...
+    integer(shortInt), dimension(:), allocatable :: perm
+    type(particleState)                          :: tmp
+    character(100), parameter :: Here = 'sortBybroodID (particleDungeon_class.f90)'
+
+    ! Count number of particles with each shower ID
+    count = 0
+    do i = 1, self % pop
+      id = self % prisoners(i) % broodID
+
+      if (id < 1 .or. id > k) call fatalError(Here, 'Shower ID out of range: '//numToChar(id))
```
valeriaRaffuzzi, line 494:
```suggestion
  !! Reorder the dungeon so the brood ID is in the ascending order
  !!
  !! Args:
  !!   k [in] -> Maximum brood ID
  !!
  subroutine sortByBroodID(self, k)
    class(particleDungeon), intent(inout)        :: self
    integer(shortInt), intent(in)                :: k
    integer(shortInt), dimension(k)              :: count
    integer(shortInt)                            :: i, id, loc, c
    integer(shortInt), dimension(:), allocatable :: perm
    type(particleState)                          :: tmp
    character(100), parameter :: Here = 'sortBybroodID (particleDungeon_class.f90)'

    ! Count number of particles with each brood ID
    count = 0
    do i = 1, self % pop
      id = self % prisoners(i) % broodID

      if (id < 1 .or. id > k) call fatalError(Here, 'Brood ID out of range: '//numToChar(id))
```

Mikolaj-A-Kowalski:
Thanks for catching the showeIds in the comments!

valeriaRaffuzzi:
> Thanks for catching the showeIds in the comments!

No problem, it was a jetlag late night activity! Feel free to merge this when the comments are fixed :)

Mikolaj-A-Kowalski:
Grep says it is fixed now. No 'shower' in SCONE. Merging

PR #117, opened 2024-02-07
Tracking xs

valeriaRaffuzzi:
This PR fixes the problems with scoring virtual collisions, and might be useful in the future when adding TMS to the mix. 

It removes the old getTransMat function from the database_inter, and introduces getTrackingXS. This is the cross section used by DT and ST to sample path length, i.e., total mat xs or majorant xs on request. The tracking xs is saved in a trackingCache, that is then used by the tallies when a virtual collision in requested. One question I actually have is about the cache. The majorant cache is initialised with dimension nMaj, but when it's read the index is hardcoded 1. I would remove the nMaj option and just make both the majorant and the tracking cache a scalar rather than an array. What do you think? 

Now it's safe to use `handleVirtual 1` in both DT and ST. 

The virtual collision treatment was also introduced in the Clerks other than collisionClerk that call reportCollisionIn. Some of those clerks needed a check for void, given that DT now sends reports also in void. 

NB: for now I think it makes sense to keep both the majorantCache and the trackingCache, because they serve a different purpose. 

I tested this with all combinations of ST, DT, HT and virtual flag 0 or 1.


NuclearData/nuclearDatabase_inter.f90
```diff
@@ -21,8 +21,8 @@ module nuclearDatabase_inter
   !! subclasses of this type.
   !!
   !! Interface:
-  !!   getTransMatXS -> returns transport Material XS given a particle
-  !!   getTotalMatXS -> returns total Material XS fiven a particle
+  !!   getTrackingXS -> returns XS used to sample track lenght
```
ChasingNeutrons, line 24: length*

Tallies/TallyClerks/collisionProbabilityClerk_class.f90
```diff
@@ -202,7 +202,7 @@ subroutine reportInColl(self, p, xsData, mem, virtual)
     ! neutrons which collide outside the mapped region of phase space
     ! These correspond to index = 0
 
-    ! Calculate collision probability 
+    ! Calculate collision probability
```
ChasingNeutrons, line 205: Thanks for finding all of these trailing spaces... I need a better editor.

ChasingNeutrons (approved):
All looks good to me. I have a mild preference for having virtual collisions on by default, but after last time that could bite us...

ChasingNeutrons (approved):

ChasingNeutrons:
On the original commit I think there was some debating about whether to have it on by default. But indeed, there are arguments either way.

PR #120, opened 2024-04-05
Some new benchmark problems

ChasingNeutrons:
In addition to the MCNP benchmarks we have, I thought we should have some more.

I changed the BenchmarksMCNP folder to Benchmarks/MCNP. I also added the OECD-NEA source convergence problem 1 and a 2D and 3D BEAVRS model.

There are a few small issues with the BEAVRS problems. Common to both is that the temperatures for the data aren't exactly as specified - I'm using 600K while the actual (isothermal) temperature is 566K. This can be treated more appropriately when we have OTF Doppler broadening and e.g. stochastic mixing for S(alpha,beta). The second fudge in the 3D problem is that the geometry definition is presently slightly wrong: all rods are removed at the moment but, actually, four assemblies on the diagonals have rods partially inserted.

Arguably, these all fall short of 'benchmark' in that there is no commonly accepted value of k, etc, but I think they will at least be useful for testing the code in future.

I'm hoping to add Congjin's HM Benchmark in too shortly! I'll also try and add C5G7.


ChasingNeutrons:
I've now added both C5G7 and Hoogenboom-Martin.

InputFiles/Benchmarks/HoogenboomMartin/HMcore
```diff
@@ -0,0 +1,575 @@
...
+                      xax {type spaceMap;  axis x;  grid lin; N 17; min -182.07; max 182.07; }
+                      yax {type spaceMap;  axis y;  grid lin; N 17; min -182.07; max 182.07; }
+                   }
+                 } 
+
+}
+
+! Uniform fission site map
+uniformFissionSites {
+                   type uniFissSitesField;
```
valeriaRaffuzzi, line 55: Is this with a uniform map?

PR #121, opened 2024-04-11
TMS

valeriaRaffuzzi:
Here is the basic implementation of TMS. The main modifications are: 

- in aceNeutronDatabase routines to get the material temperature majorant and relative energy macroscopic xss
- few entries were added to the nuclide and material cache: nuclide -> store total temperature majorant xs and parameters at which it was calculated. material -> energy relative to which the relative macroxss are calculated, and anther set of xss for the relative energy ones
- in ceNeutronMaterial, the routine to sample the nuclide now needs to check whether to accept the sampled nuclide or not
- in collisionProcessor_inter, the outcome of the nuclide sampling can be MT = noInteraction (rejection)
- in the collision processors the relative energy for which a nuclide is accepted is stored and used to sample reaction, to calculate number of fission neutrons and to sample inelastic scattering MT. QUESTION: should the sampled energy be used for scatterInLab and ScatterFromFixed too????????
- routines in scatteringKernels to sample the relative energy and calculate the DOppler correction factor for low energies
- in ceNeutronMaterial, sampleFission also had to be adapted
- the benchmarks were adaped to run with TMS
- in all the clerks that support reportInColl, virtual collisions are on by default (this is written in the User Manual)

NOTE: TMS is applied to the materials where 'temp' is specified in the input file, and outside S(a,b) and ures energy range with a margin such that the relative energy doesn't fall into those ranges
NOTE2: in aceDB % init, there's a check with FatalError if the material temperature specified is smaller or equal a nuclide temperature

Other modifications:
- initMajorant was slightly simplified
- routines in aceNeutronNuclide to verify the ranges of Sab and Ures

NOTE3: This implementation might still be suboptimal in terms of speed. For example, trying to use TMS on water (silly anyway), when deltaT > 50K the code slows down a lot because of the massive thermal temprature majorant and lots of rejections. Hopefully this can be fixed in the future.


InputFiles/Benchmarks/BEAVRS/BEAVRS2D
```diff
@@ -1275,304 +1275,304 @@ nuclearData {
...
+      18036.03 7.8730E-09;
+      18038.03 1.4844E-09;
+      18040.03 2.3506E-06;
+      6012.03  6.7539E-08;
+      //6013.03  7.5658E-10;
+      7014.03  1.9680E-04;
+      7015.03  7.2354E-07;
+      8016.03  5.2866E-05;
+      8017.03  2.0084E-08;
+      //8018.03  1.0301E-07;
```
valeriaRaffuzzi, line 1287:
```suggestion
      //8018.03  1.0601E-07; 

```

ChasingNeutrons:
One initial comment: do we want having a temperature in a material definition to activate TMS? I think there is something to be said for having a separate TMS flag in addition to temperature, as we essentially have for S(a,b). When we come to doing multiphysics etc, temperature seems like a natural variable to have around even if there are places we don't want to apply TMS.

CollisionOperator/CollisionProcessors/collisionProcessor_inter.f90
```diff
@@ -113,25 +113,35 @@ subroutine collide(self, p, tally, thisCycle, nextCycle)
...
+    ! Choose collision nuclide and general type (Scatter, Capture or Fission)
+    call self % sampleCollision(p, tally, collDat, thisCycle, nextCycle)
+
+    ! In case of a TMS rejection, set collision as virtual
+    if (collDat % MT == noInteraction) then
+      virtual = .true.
+      thisCollision = 0
+    else
+      virtual = .false.
+      thisCollision = 1
```
ChasingNeutrons, line 132: Pedantic, but this variable isn't very clear to me. I understand reading below that it's for counting collisions. I might propose a more intuitive name like 'isCollision' or 'collisionWorth'.

CollisionOperator/CollisionProcessors/neutronCEimp_class.f90
```diff
@@ -121,8 +121,12 @@ module neutronCEimp_class
     logical(defBool)  :: implicitSites ! Generates fission sites on every fissile collision
     logical(defBool)  :: uniFissSites
 
+    ! Variance reduction requirements
     type(weightWindowsField), pointer :: weightWindowsMap
 
+    ! Useful to store
+    real(defReal) :: eSample
```
ChasingNeutrons, line 128: Maybe describe what this variable does/is here too?

CollisionOperator/CollisionProcessors/neutronCEimp_class.f90
```diff
@@ -246,13 +250,19 @@ subroutine sampleCollision(self, p, tally, collDat, thisCycle, nextCycle)
     if(.not.associated(self % mat)) call fatalError(Here, 'Material is not ceNeutronMaterial')
 
     ! Select collision nuclide
-    collDat % nucIdx = self % mat % sampleNuclide(p % E, p % pRNG)
+    call self % mat % sampleNuclide(p % E, p % pRNG, collDat % nucIdx, self % eSample)
```
ChasingNeutrons, line 253: On eSample, is it more natural to have it in the collisionProcessor than the particle? I ask with respect to (someday!) not having to make thread private copies of the processor.

valeriaRaffuzzi: Yes that's a fair point. Adding a property like eCollision to the particle would be also ok!

CollisionOperator/CollisionProcessors/neutronCEstd_class.f90
```diff
@@ -50,11 +52,11 @@ module neutronCEstd_class
   !!  minE    -> minimum energy cut-off [MeV] (default = 1.0E-11)
   !!  maxE    -> maximum energy. Higher energies are set to maximum (not re-rolled) [MeV]
   !!             (default = 20.0)
-  !!  thresh_E -> Energy threshold for explicit treatment of target nuclide movement [-].
-  !!              Target movment is sampled if neutron energy E < kT * thresh_E where
+  !!  threshE -> Energy threshold for explicit treatment of target nuclide movement [-].
+  !!              Target movment is sampled if neutron energy E < kT * threshE where
```
ChasingNeutrons, line 56: same typos in movement* and treatment* below

CollisionOperator/CollisionProcessors/neutronCEstd_class.f90
```diff
@@ -77,11 +79,14 @@ module neutronCEstd_class
...
     real(defReal) :: maxE
-    real(defReal) :: thresh_E
-    real(defReal) :: thresh_A
+    real(defReal) :: threshE
+    real(defReal) :: threshA
     real(defReal) :: DBRCeMin
     real(defReal) :: DBRCeMax
 
+    ! Useful to store
+    real(defReal) :: eSample
```
ChasingNeutrons, line 88: as before, if it remains a property.

CollisionOperator/CollisionProcessors/neutronCEstd_class.f90
```diff
@@ -294,19 +308,28 @@ subroutine elastic(self, p, tally, collDat, thisCycle, nextCycle)
...
     ! Get reaction
     reac => uncorrelatedReactionCE_CptrCast( self % xsData % getReaction(collDat % MT, collDat % nucIdx))
-    if(.not.associated(reac)) call fatalError(Here,'Failed to get elastic neutron scatter')
+    if (.not.associated(reac)) call fatalError(Here,'Failed to get elastic neutron scatter')
 
     ! Scatter particle
     collDat % A =  self % nuc % getMass()
-    collDat % kT = self % nuc % getkT()
+
+    ! Retrieve kT from either material or nuclide
```
ChasingNeutrons, line 321: This is mostly about my ignorance and probably not something to address now...
Is it conceivable to include only some nuclides in TMS in a given material? E.g., we could just rule out doing TMS on hydrogen if so.

CollisionOperator/scatteringKernels_func.f90
```diff
@@ -113,7 +115,7 @@ function targetVelocity_constXS(E,dir,A,kT,rand) result (V_t)
     Y = sqrt(A*E/kT)
 
     ! Calculate threshold factor alpha
-    alpha = 2.0/(Y*sqrt(PI)+2.0)
+    alpha = TWO/(Y*sqrt(PI) + TWO)
```
ChasingNeutrons, line 118: make sqrt(PI) a universal variable? One would imagine it must turn up elsewhere...

CollisionOperator/scatteringKernels_func.f90
```diff
@@ -127,7 +129,7 @@ function targetVelocity_constXS(E,dir,A,kT,rand) result (V_t)
 
     ! Calculate azimithal angle for target and obtain target direction
```
ChasingNeutrons, line 130: azimuthal*

CollisionOperator/scatteringKernels_func.f90
```diff
@@ -202,6 +204,71 @@ function targetVelocity_DBRCXS(nuc, E, dir, A, kT, rand, tempMaj) result (V_t)
...
+    alpha = sqrt(A * E / kT)
+
+    invAlph = ONE / alpha
+
+    if (alpha > 250) then
+      g = ONE
+    else if (alpha > 2.568) then
+      g = ONE + HALF * invAlph * invAlph
+    else
+      g = (ONE + HALF * invAlph * invAlph) * ERF(alpha) + EXP(-alpha * alpha) * invAlph / sqrt(PI)
```
ChasingNeutrons, line 268: exp* for consistency

InputFiles/Benchmarks/BEAVRS/BEAVRS
```diff
@@ -1956,304 +1956,304 @@ nuclearData {
     Air {
       temp 566;
```
ChasingNeutrons, line 1957: For air, I'm inclined to leave it as it was, just because it won't have a big neutronic effect and it feels a bit ovekill to Doppler broaden it 36K down from the library. This may be completely immaterial though.

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
```diff
@@ -291,6 +299,47 @@ subroutine energyBounds(self, E_min, E_max)
...
+
+      end do
+
+    end associate
+
+  end subroutine updateTotalTempMajXS
+
   !!
   !! Make sure that totalXS of material with matIdx is at energy E
   !! in ceNeutronChache
```
ChasingNeutrons, line 345: Typo while here. ceNeutronCache*

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
```diff
@@ -399,32 +455,116 @@ subroutine updateMacroXSs(self, E, matIdx, rand)
...
+  !!
+  !! Subroutine to update the macroscopic cross sections in a given material
+  !! at given temperature, sampling a relative energy per each nuclide and applying
+  !! a Doppler correction factor
+  !!
+  !! Args:
+  !!   E [in]         -> Incident neutron energy for which the relative energy xss are found
+  !!   matIdx [in]    -> Index of material for which the relative energy xss are found
+  !!
+  subroutine updateRelEnMacroXSs(self, E, matIdx, rand)
```
ChasingNeutrons, line 502: I have a mild preference for the name to be RelE rather than RelEn, just for consistency with other things.

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
```diff
@@ -399,32 +455,116 @@ subroutine updateMacroXSs(self, E, matIdx, rand)
...
+            ! Add microscopic XSs
+            call matCache % xssRel % add(nucCache % xss, dens * doppCorr)
+
+          end associate
+
+        end do
+
+      end if
+
+      ! Update cache, and ensure that
```
ChasingNeutrons, line 559: Unfinished comment?

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
```diff
@@ -399,32 +455,116 @@ subroutine updateMacroXSs(self, E, matIdx, rand)
...
+      matCache % E_tot  = ZERO
+      matCache % E_tail = ZERO
+
+    end associate
+
+  end subroutine updateRelEnMacroXSs
+
   !!
   !! Make sure that totalXS of nuclide with nucIdx is at energy E
   !! in ceNeutronChache
```
ChasingNeutrons, line 570: typo

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
```diff
@@ -660,42 +851,100 @@ subroutine init(self, dict, ptr, silent )
       i = nucSet % next(i)
     end do
 
+    ! Calculate energy bounds
+    self % Ebounds(1) = self % nuclides(1) % eGrid(1)
```
ChasingNeutrons, line 855: Rename Ebounds to eBounds for consistency?

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
```diff
@@ -660,42 +851,100 @@ subroutine init(self, dict, ptr, silent )
...
+    end do
+
     ! Build Material definitions
     allocate(self % materials(mm_nMat()))
     allocate(nucIdxs(maxNuc))
     do i = 1, mm_nMat()
       mat => mm_getMatPtr(i)
 
       ! Load nuclide indices on storage space
+      ! Find if material if fissile and useful energy limits
```
ChasingNeutrons, line 872: Where is the useful energy limits bit? I think this comment may be outdated.

NuclearData/ceNeutronData/aceDatabase/aceNeutronNuclide_class.f90
```diff
@@ -589,45 +587,112 @@ end subroutine getUrrXSs
...
       idx = idx + 1
 
       ! Find XS and energy at index
-      xs = self % mainData(ESCATTER_XS, idx)
-      E = self % eGrid(idx)
+      xs = self % mainData(reaction, idx)
+      E  = self % eGrid(idx)
 
       ! Compare cross sections and possibly update majorant
       if (xs > maj) then
```
ChasingNeutrons, line 635: Maybe slightly neater: maj = max(maj,xs)?

NuclearData/ceNeutronData/ceNeutronCache_mod.f90
```diff
@@ -116,41 +121,30 @@ module ceNeutronCache_mod
   !! Args:
   !!   nMat [in]  -> Number of materials
   !!   nNuc [in]  -> Number of nuclides
-  !!   nMaj [in]  -> Optional. Number of majorant Xss (Default = 1)
```
ChasingNeutrons, line 119: Are we sure we want to get rid of this? Maybe it would have been used for, e.g., holes? I suppose @Mikolaj-A-Kowalski may have some thoughts.

NuclearData/ceNeutronData/ceNeutronDatabase_inter.f90
```diff
@@ -196,7 +197,31 @@ subroutine updateMicroXSs(self, E, nucIdx, rand)
...
+    !!
+    !! Args:
+    !!   E [in]      -> required energy [MeV]
+    !!   kT [in]     -> Thermal energy of TMS material
+    !!   nucIdx [in] -> material index that needs to be updated
+    !!
+    !! Errors:
+    !!   FatalError if material kT is smaller than the nuclide kT
+    !!
+    subroutine updateTotalTempNucXS(self, E, kT, nucIdx)
```
ChasingNeutrons, line 215: Can this be given a more descriptive name? I suppose for me Temp has an annoying double connotation. Reading the comments, it isn't clear how the name relates to the function other than TMS - maybe updateTotalTMSNucXS?

NuclearData/ceNeutronData/ceNeutronMaterial_class.f90
```diff
@@ -129,13 +144,13 @@ subroutine setComposition(self, dens, nucIdxs)
     character(100), parameter :: Here = 'setComposition (ceNeutronMaterial_class.f90)'
 
     ! Check input
-    if(size(dens) /= size(nucIdxs)) call fatalError(Here,'Diffrent sizes of density and nuclide vector')
-    if(any(dens < ZERO)) call fatalError(Here,'-ve nuclide densities are present')
-    if(size(dens) == 0)  call fatalError(Here,'Empty composition is not allowed')
+    if (size(dens) /= size(nucIdxs)) call fatalError(Here,'Diffrent sizes of density and nuclide vector')
```
ChasingNeutrons, line 147: Different*

NuclearData/ceNeutronData/ceNeutronMaterial_class.f90
```diff
@@ -215,6 +242,28 @@ elemental function isFissile(self) result(isIt)
...
 
+  !!
+  !! Return .true. if TMS is on in the material and the provided energy is not
+  !! within the ures or S(a,b) range in the material
+  !!
+  !! Args:
+  !!   E [in] -> test energy
+  !!
+  !! Result:
+  !!   .TRUE. if conditions for using TMS are satisfied, .FALSE. otherwise
```
ChasingNeutrons, line 253: Consistent capitalisation of bools

NuclearData/ceNeutronData/ceNeutronMaterial_class.f90
```diff
@@ -224,48 +273,118 @@ end function isFissile
   !! Args:
   !!   E [in]       -> incident energy [MeV]
   !!   rand [inout] -> random number generator
-  !!
-  !! Result:
-  !!   nucIdx of the sampled nuclide for collision
+  !!   nucIdx [out] -> sampler nuclide index
```
ChasingNeutrons, line 276: sampled*

NuclearData/ceNeutronData/ceNeutronMaterial_class.f90
```diff
@@ -224,48 +273,118 @@ end function isFissile
...
     ! Print error message as the inversion failed
     call fatalError(Here,'Nuclide sampling loop failed to terminate')
 
-  end function sampleNuclide
+  end subroutine sampleNuclide
 
   !!
   !! Sample fission nuclide given that a fission neutron was produced
   !!
   !! Basicly samples from P(nucIdx| fission neutron produced in material)
```
ChasingNeutrons, line 386: basically*

NuclearData/ceNeutronData/ceNeutronMaterial_class.f90
```diff
@@ -287,30 +406,53 @@ function sampleFission(self, E, rand) result(nucIdx)
...
+
       nucIdx = self % nuclides(i)
-      if(E /= nuclideCache(nucIdx) % E_tail) call self % data % updateMicroXSs(E, nucIdx, rand)
-      xs = xs - nuclideCache(nucIdx) % xss % nuFission * self % dens(i)
-      if(xs < ZERO) return
+
+      ! If the material uses TMS, the Doppler correction factor is needed
+      if (self % useTMS(E)) then
+
+        ! If the material is using TMS, the Doppler correction factor is needed
```
ChasingNeutrons, line 436: duplicate comment

NuclearData/ceNeutronData/ceNeutronMaterial_class.f90
```diff
@@ -287,30 +406,53 @@ function sampleFission(self, E, rand) result(nucIdx)
...
+        nuc => ceNeutronNuclide_CptrCast(self % data % getNuclide(nucIdx))
+        if (.not. associated(nuc)) call fatalError(Here, 'Failed to retive CE Neutron Nuclide')
```
ChasingNeutrons, line 438: retrieve*

NuclearData/ceNeutronData/ceNeutronNuclide_inter.f90
```diff
@@ -153,7 +154,26 @@ function elScatteringXS(self, E) result(xs)
 
     end function elScatteringXS
 
+    !!
+    !! Function that checks whether this nuclide at the provided energy should
+    !! has S(a,b) elastic scattering data or not
```
ChasingNeutrons, line 159: should have*

NuclearData/materialMenu_mod.f90
```diff
@@ -100,6 +100,7 @@ module materialMenu_mod
     real(defReal),dimension(:),allocatable     :: dens
     type(nuclideInfo),dimension(:),allocatable :: nuclides
     type(dictionary)                           :: extraInfo
+    logical(defBool)                           :: hasTMS = .false.
```
ChasingNeutrons, line 103: I would be inclined to keep TMS out of the material menu if possible. For example, if we wanted to do multipole or if we wanted to do MG data. Maybe there is no other obvious way? I suppose we are in the same difficulty for Sab too. This also comes back to temperature being something that is desirable for any generic material to keep track of and ideally not something that automatically triggers TMS ( MG could benefit from having a temperature despite not using TMS).

NuclearData/materialMenu_mod.f90
```diff
@@ -285,10 +288,13 @@ subroutine init_materialItem(self, name, idx, dict)
...
     call self % kill()
 
-    ! Load easy components c
+    ! Load easy components properties
     self % name = name
     self % matIdx = idx
-    call dict % get(self % T,'temp')
+
+    ! If material temperature is present TMS is switched on
+    if (dict % isPresent('temp')) self % hasTMS = .true.
```
ChasingNeutrons, line 296: If all of the above comment can't be addressed, even just having a separate flag for TMS and temperature would be an improvement for future developments I think.

SharedModules/universalVariables.f90
```diff
@@ -64,19 +64,21 @@ module universalVariables
                                   P_NEUTRON_MG = 2
 
   ! Search error codes
-  integer(shortInt), parameter :: valueOutsideArray = -1,&
-                                  tooManyIter       = -2,&
+  integer(shortInt), parameter :: valueOutsideArray = -1, &
+                                  tooManyIter       = -2, &
                                   targetNotFound    = -3, &
-                                  NOT_FOUND         = -3
+                                  NOT_FOUND         = -3, &
```
ChasingNeutrons, line 70: why have separate flags for targetNotFound and NOT_FOUND?

SharedModules/universalVariables.f90
```diff
@@ -64,19 +64,21 @@ module universalVariables
...
   ! Integer indexes for type of tracking cross section requested
   integer(shortInt), parameter :: MATERIAL_XS = 1, &
                                   MAJORANT_XS = 2, &
                                   TRACKING_XS = 3
 
   ! Physical constants
-  real(defReal), parameter :: neutronMass = 939.5654133_defReal, &   ! Neutron mass in MeV/c^2
+  real(defReal), parameter :: neutronMass = 939.5654133_defReal,   & ! Neutron mass in MeV/c^2
                               lightSpeed  = 2.99792458e10_defReal, & ! Light speed in cm/s
+                              kBoltzmann  = 1.380649e-23_defReal,  & ! Bolztman constant in J/K
```
ChasingNeutrons, line 81: Boltzmann* in comment

Tallies/TallyClerks/collisionClerk_class.f90
```diff
@@ -194,29 +196,25 @@ subroutine reportInColl(self, p, xsData, mem, virtual)
     logical(defBool), intent(in)          :: virtual
     type(particleState)                   :: state
     integer(shortInt)                     :: binIdx, i
+    type(neutronMacroXSs)                 :: xss
+    class(neutronMaterial), pointer       :: mat
     integer(longInt)                      :: adrr
```
ChasingNeutrons, line 201: While here, it should be addr*, not adrr

Tallies/TallyClerks/collisionClerk_class.f90
```diff
@@ -194,29 +196,25 @@ subroutine reportInColl(self, p, xsData, mem, virtual)
...
 
     ! Return if collision is virtual but virtual collision handling is off
-    if (self % virtual) then
-      ! Retrieve tracking cross section from cache
-      flux = p % w / xsData % getTrackingXS(p, p % matIdx(), TRACKING_XS)
-    else
-      if (virtual) return
-      flux = p % w / xsData % getTotalMatXS(p, p % matIdx())
-    end if
+    if ((.not. self % virtual) .and. virtual) return
```
ChasingNeutrons, line 206: Maybe change the property name? This is slightly confusing to read. Perhaps handleVirtual?

Tallies/TallyClerks/keffImplicitClerk_class.f90
```diff
@@ -33,7 +33,6 @@ module keffImplicitClerk_class
                                   IMP_ABS      = 2 ,&  ! Implicit neutron absorbtion
                                   ANA_LEAK     = 3 ,&  ! Analog Leakage
                                   K_EFF        = 4     ! k-eff estimate
-
   !!
   !! A simple implicit k-eff estimator based on collison estimator of reaction rates,
   !! and an analog estimators of (N,XN) reactions and leakage
```
ChasingNeutrons, line 38: and on* analog

Tallies/TallyClerks/simpleFMClerk_class.f90
```diff
@@ -194,21 +196,16 @@ subroutine reportInColl(self, p, xsData, mem, virtual)
...
     type(scoreMemory), intent(inout)     :: mem
     logical(defBool), intent(in)         :: virtual
+    type(neutronMacroXSs)                :: xss
+    class(neutronMaterial), pointer      :: mat
     type(particleState)                  :: state
     integer(shortInt)                    :: sIdx, cIdx
     integer(longInt)                     :: addr
     real(defReal)                        :: score, flux
-    class(neutronMaterial), pointer      :: mat
     character(100), parameter :: Here = 'reportInColl simpleFMClear_class.f90'
```
ChasingNeutrons, line 205: typoe: simpleFMClerk*

Tallies/TallyClerks/simpleFMClerk_class.f90
```diff
@@ -254,18 +263,18 @@ subroutine reportCycleEnd(self, end, mem)
     integer(longInt)                    :: addrFM
     real(defReal)                       :: normFactor
 
-    if(mem % lastCycle()) then
+    if (mem % lastCycle()) then
       ! Set address to the start of Fission Matrix
       ! Decrease by 1 to get correct addres on the fisrt iteration of the loop
```
ChasingNeutrons, line 268: address* of the first*

ChasingNeutrons (commented):
Thanks for this. My most major comment is just on keeping temperature as a separate variable which I think we would like to distinguish without demanding TMS (e.g., for multigroup calculations where we might eventually have multiphysics).

docs/User Manual.rst
```diff
@@ -780,15 +780,16 @@ The *materials* definition is structured as: ::
...
-  At the moment ``temp`` is not used in any way since SCONE has no way to treat
-  the temperature dependence of cross-sections. It is included for future use.
-  To change the temperature, a user needs to set appropriate suffix to each
-  individual nuclide in the composition definition.
+  The temperature specified by ``temp`` must be higher than the temperatures of the 
+  nuclides in the material composition.
 
+.. note::
+  *IMPORTANT*: When using TMS, all the tallies based on the collision estimator have to
+  allow scoring virtual collisions, otherwise the results will be biased
```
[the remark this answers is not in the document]
valeriaRaffuzzi: Track length is also biased. I forgot to mention it

Tallies/TallyClerks/collisionClerk_class.f90
```diff
@@ -225,6 +223,25 @@ subroutine reportInColl(self, p, xsData, mem, virtual)
...
+      ! Get material pointer
+      mat => neutronMaterial_CptrCast(xsData % getMaterial(p % matIdx()))
+      if (.not.associated(mat)) then
+        call fatalError(Here,'Unrecognised type of material was retrived from nuclearDatabase')
+      end if
+
+      ! Obtain xss
+      call mat % getMacroXSs(xss, p)
+
+      flux = p % w / xss % total
```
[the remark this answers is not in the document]
valeriaRaffuzzi: Not really, because in TMS getTotalMatXS gets you the material temperature majorant, while getMacroXSs retrieves the material macroscopic xss at the relative energy. I should probably comment the code better to emphasize this.

valeriaRaffuzzi:
> Thanks for this. My most major comment is just on keeping temperature as a separate variable which I think we would like to distinguish without demanding TMS (e.g., for multigroup calculations where we might eventually have multiphysics).

Thanks for reviewing! I am ok with changing keyword to, say, 'tms' instead of 'temp'. But I don't think we can easily skip the materialMenu step, not I think it is really an issue to go through it!

CollisionOperator/CollisionProcessors/collisionProcessor_inter.f90
valeriaRaffuzzi: collisionCount? I don't like collisionWorth or isCollision. I actually didn't mind thisCollision

ChasingNeutrons: addCollision perhaps?

CollisionOperator/CollisionProcessors/neutronCEstd_class.f90
valeriaRaffuzzi: It would be possible I think! One could store some flag in the nuclide and use the normal xs instead of the temperature majorant for that one

NuclearData/ceNeutronData/ceNeutronCache_mod.f90
valeriaRaffuzzi: I don't have strong feelings, I just thought the code looked neater. Happy to change it back if you both prefer that way.

NuclearData/ceNeutronData/aceDatabase/aceNeutronNuclide_class.f90
valeriaRaffuzzi: I don't know how I missed that

NuclearData/materialMenu_mod.f90
valeriaRaffuzzi: I agree with changing keywords, but I don't think it's such a problem to go through materialMenu. That's how we read everything about the materials. That info is not used during transport because it is stored inside ceNeutronMaterial.

SharedModules/universalVariables.f90
valeriaRaffuzzi: Good q

valeriaRaffuzzi: Actually, aesthetics in the code probably!!

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
valeriaRaffuzzi: I don't like many capitals in a row in a name! Like updateRelEMacroXSs. The 'E' gets lost in there.

Mikolaj-A-Kowalski (commented):
Submitting the review but I didn't manage to finish before the changes. 
Will go over it again.

CollisionOperator/scatteringKernels_func.f90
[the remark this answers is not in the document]
ChasingNeutrons: On this, might there be some value in precomputing 1/sqrt(PI)? Very minor optimisation.

ParticleObjects/particle_class.f90
```diff
@@ -56,6 +57,7 @@ module particle_class
     integer(shortInt)          :: cellIdx  = -1     ! Cell idx at the lowest coord level
     integer(shortInt)          :: uniqueID = -1     ! Unique id at the lowest coord level
     integer(shortInt)          :: collisionN = 0    ! Number of collisions
+    real(defReal)              :: collisionE = ZERO ! Collision energy
```
Mikolaj-A-Kowalski, line 60: Is it possible to keep the `collisionE` out of the `particleState`? Or is it necessary there to work with some tally map interfaces and such?

ParticleObjects/particle_class.f90
```diff
@@ -111,6 +114,7 @@ module particle_class
     integer(shortInt)          :: fate = 0       ! Neutron's fate after being subjected to an operator
     integer(shortInt)          :: type           ! Particle type
     integer(shortInt)          :: collisionN = 0 ! Index of the number of collisions the particle went through
+    real(defReal)              :: collisionE = ZERO ! Collision energy
```
Mikolaj-A-Kowalski, line 117: I am not fully sure about putting the `collisionE` in the particle... 
For me the problem is that once the particle leaves a collision processor this information is missing necessary context: temperature and nuclide)

Tallies/TallyClerks/collisionClerk_class.f90
```diff
@@ -120,7 +122,7 @@ subroutine init(self, dict, name)
     self % width = size(responseNames)
 
     ! Handle virtual collisions
-    call dict % getOrDefault(self % virtual,'handleVirtual', .false.)
+    call dict % getOrDefault(self % handleVirtual,'handleVirtual', .true.)
```
Mikolaj-A-Kowalski, line 125:
```suggestion
    call dict % getOrDefault(self % handleVirtual, 'handleVirtual', .true.)
```
?

Tallies/TallyClerks/collisionClerk_class.f90
```diff
@@ -225,13 +223,32 @@ subroutine reportInColl(self, p, xsData, mem, virtual)
     ! Return if invalid bin index
     if (binIdx == 0) return
 
+    ! Return if collision is virtual but virtual collision handling is off
```
Mikolaj-A-Kowalski, line 226:
```suggestion

```
The comment seems to be dead. This handling was moved up the function?

Tallies/TallyClerks/collisionClerk_class.f90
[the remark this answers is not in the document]
Mikolaj-A-Kowalski: I think this point should be elaborated in the doc comment of the `getTotalMatXS` and `getMacroXSs`. It is a bit difficult to understand what cross-sections are obtained bu which call... however I am not sure how to make it cleaner at this moment.

CollisionOperator/CollisionProcessors/collisionProcessor_inter.f90
valeriaRaffuzzi: collisionIncrement is my fav

CollisionOperator/CollisionProcessors/neutronCEimp_class.f90
valeriaRaffuzzi: But collDat is also a property of the processor right? So the issue about the need for a thread private processor would stay. 
I think it would be nicer to keep that energy within the collision processor, but the implementation with the particle should be safe for now because I expect that property to be only read and updated in the collision processor. I can't think of an ultimate solution here I think.

Tallies/TallyClerks/collisionClerk_class.f90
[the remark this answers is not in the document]
Mikolaj-A-Kowalski: Perhaps there is no way around introducing some kind of `getTrackingMatXs`? This double meaning between the `getMacroXS` and `getTotalMatXS` is going to cause confusion and bug down the line I fear. 

All of this obviously gets a bit confusing since the cross-sections have now become random variables. However, this confusion is not specific to TMS since it appears in URR as well. Basically we need to somehow separate the 'sampled' XSs at a particular state of the flight from a deterministic supremum on their values  (`TrackingXS`?)

Mikolaj-A-Kowalski (commented):
I think the biggest problem to address is to clarify the nuclear data interface in the presence of random cross-sections. 
Basically we need to say clearly which functions return maximum value of the cross-section and which return a value obtained by a sample. I have a suspicion that we can call the maxima 'tracking' cross-section and in all other contexts functions would return sampled values? But I might be missing some problems. 

I also believe having a paragraph to describe how we handle the random cross-sections (in the docs) would be really helpful, because I have a feeling it is difficult to make sense of what is going on from the code alone.

Tallies/TallyClerks/mgXsClerk_class.f90
```diff
@@ -228,13 +231,10 @@ subroutine reportInColl(self, p, xsData, mem, virtual)
...
-      ! Retrieve tracking cross section from cache
-      flux = p % w / xsData % getTrackingXS(p, p % matIdx(), TRACKING_XS)
-    else
-      if (virtual) return
-      flux = p % w / xsData % getTotalMatXS(p, p % matIdx())
-    end if
+    if ((.not. self % virtual) .and. virtual) return
+
+    ! Ensure we're not in void (could happen when scoring virtual collisions)
+    if (p % matIdx() == VOID_MAT) return
```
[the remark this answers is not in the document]
valeriaRaffuzzi: I added that check to avoid trying to look up cross sections in void. Now, for example, I removed this check from the collisionClerk (so that one is able to calculate the flux in void) but I added it to macroResponse to avoid look ups in void. 

I guess in the mgXsClerk I could score a flux in void but set all the other xss to zero, that would make it correct. I can have a look at that in this PR, I don't see why postponing!

NuclearData/ceNeutronData/ceNeutronNuclide_inter.f90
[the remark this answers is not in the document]
valeriaRaffuzzi: Not only. This is called in nucDatabase % updateTotNucXS and nucDatabase % updateMicroNucXS.

This check before was included inside nucDatabase % updateTotNucXS and stored in cache, but that was becoming a dangerous game. The cache logic was too complex and error prone (I think in fact this part was giving me trouble when looking up nuclide relative energy xss for TMS) so I prefer to have this check indipendent from the cache and repeat it whenever it's needed.

NuclearData/ceNeutronData/ceNeutronMaterial_class.f90
```diff
@@ -147,25 +162,37 @@ end subroutine setComposition
...
+    real(defReal), intent(in), optional                     :: eUpperSab
+    real(defReal), intent(in), optional                     :: eLowerURR
+
+    if (present(database))  self % data    => database
+    if (present(fissile))   self % fissile = fissile
+    if (present(matIdx))    self % matIdx  = matIdx
+    if (present(hasTMS))    self % hasTMS  = hasTMS
+    if (present(temp))      self % kT = (kBoltzmann * temp) / joulesPerMeV
+    if (present(eUpperSab)) self % eUpperSab  = eUpperSab
+    if (present(eLowerURR)) self % eLowerURR  = eLowerURR
```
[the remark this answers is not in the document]
valeriaRaffuzzi: Fair! We can ensure they are positive and that eUpperSab < eLowerURR.

NuclearData/ceNeutronData/ceNeutronMaterial_class.f90
```diff
@@ -48,10 +53,14 @@ module ceNeutronMaterial_class
...
     integer(shortInt)                            :: matIdx = 0
+    real(defReal)                                :: kT     = ZERO
     class(ceNeutronDatabase), pointer            :: data => null()
     real(defReal), dimension(:), allocatable     :: dens
     integer(shortInt), dimension(:), allocatable :: nuclides
-    logical(defBool)                             :: fissile =.false.
+    logical(defBool)                             :: fissile = .false.
+    logical(defBool)                             :: hasTMS  = .false.
+    real(defReal)                                :: eUpperSab = ZERO
+    real(defReal)                                :: eLowerURR = ZERO
```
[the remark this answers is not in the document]
valeriaRaffuzzi: I calculate those energy limits for each material independently by looping over all nuclides and picking the most restrictive energies. 

Maybe this is over conservative and one should check the energies of each nuclide during transport, and only broaden the nuclides that pass this check? That's probably more rigorous.

NuclearData/ceNeutronData/ceNeutronMaterial_class.f90
```diff
@@ -224,48 +273,118 @@ end function isFissile
   !! Args:
   !!   E [in]       -> incident energy [MeV]
   !!   rand [inout] -> random number generator
-  !!
-  !! Result:
-  !!   nucIdx of the sampled nuclide for collision
+  !!   nucIdx [out] -> sampler nuclide index
+  !!   eOut [out]   -> energy used for the TMS rejection sampling
```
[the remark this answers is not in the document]
valeriaRaffuzzi: Yeah much better. I'll include that in the next round of reviews (gitHub commit suggestion is convenient but then we end up with tens of commits, so I'll just keep it clean now that it's easy)

NuclearData/ceNeutronData/ceNeutronCache_mod.f90
```diff
@@ -52,22 +57,22 @@ module ceNeutronCache_mod
   !!   f      -> Interpolation factor for the nuclide at energy E_tot
   !!   idx    -> Index on a nuclide grid for energy E_tot
   !!   xss    -> Cached Cross-Sections values
-  !!   needsSabInel -> Flag that tells if the nuclide is using thermal inelastic
-  !!                   scattering data
-  !!   needsSabEl   -> Flag that tells if the nuclide is using thermal elastic
-  !!                   scattering data
-  !!   needsUrr     -> Flag that tells if the nuclide is using unresolved resonance
-  !!                   probability tables
+  !!   E_maj     -> Energy at which the nuclide temperature majorant xs is stored (for TMS)
```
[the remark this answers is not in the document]
valeriaRaffuzzi: Yes it should be the same. But occasionally during transport one might have to retrieve relative energy xss, so the two energy flags could be different. 

Given that the temperature majorant and the other xss are updated in different places, using the same energy flag would result in errors for sure.

NuclearData/ceNeutronData/ceNeutronCache_mod.f90
```diff
@@ -33,14 +33,19 @@ module ceNeutronCache_mod
   !!   E_tail -> Energy of all XSs in xss except total
   !!   f      -> Interpolation factor for the nuclide at energy E_tot
   !!   idx    -> Index on a nuclide grid for energy E_tot
-  !!   xss    -> Cached Cross-Section values
+  !!   xss    -> Cached cross-section values
+  !!   E_rel  -> Base energy for which relative energy cross sections are found (for TMS)
+  !!   xssRel -> Cached effective cross-section values at energy relative to E_rel (for TMS)
```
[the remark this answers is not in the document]
valeriaRaffuzzi: > The XSs in `xssRel` are cross-sections that contain contributions from other nuclides in the mixture at different relative energies?

This is exactly it. I wanted to store those instead of resampling for when you need to score multiple tallies per each collision. One could resample and that would be equally correct, but much slower. So it depends on what we prioritise.

So E_rel is again the energy of the particle in lab frame, but those xssRel are different from the material xss(E_rel). So again having two flags is necessary.

NuclearData/ceNeutronData/ceNeutronCache_mod.f90
```diff
@@ -52,22 +57,22 @@ module ceNeutronCache_mod
...
-  !!   needsSabInel -> Flag that tells if the nuclide is using thermal inelastic
-  !!                   scattering data
-  !!   needsSabEl   -> Flag that tells if the nuclide is using thermal elastic
-  !!                   scattering data
-  !!   needsUrr     -> Flag that tells if the nuclide is using unresolved resonance
-  !!                   probability tables
+  !!   E_maj     -> Energy at which the nuclide temperature majorant xs is stored (for TMS)
+  !!   deltakT   -> Difference between TMS material and nuclide thermal energy (for TMS)
+  !!   tempMajXS -> Temperature majorant xs value (for TMS)
+  !!   doppCorr  -> Doppler correction factor value (for TMS)
```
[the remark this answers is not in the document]
valeriaRaffuzzi: No I don't think so. The thing is that we don't explain the physics or the algorithms in the User Manual, otherwise that would become messy. Maybe the best place to add an explanation is inside scatteringKernels where doppCorr is calculated?

NuclearData/ceNeutronData/ceNeutronMaterial_class.f90
```diff
@@ -215,6 +242,28 @@ elemental function isFissile(self) result(isIt)
...
+  !! Args:
+  !!   E [in] -> test energy
+  !!
+  !! Result:
+  !!   .TRUE. if conditions for using TMS are satisfied, .FALSE. otherwise
+  !!
+  !! Errors:
+  !!   None
+  !!
+  elemental function useTMS(self, E) result(shouldIt)
```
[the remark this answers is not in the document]
valeriaRaffuzzi: Is there any other OTF Doppler broadening that uses the same machinery as TMS? 

I think I would leave it as is, because this check is used during transport when performing operations that I believe are specific to TMS. Like in ceNeutronMaterial % sampleCollision, there's the TMS route and the normal route. Or when looking up xss, whether you need the total material or the temperature majorant is determined by this check. And I don't how general this really is, but I might be missing something.

NuclearData/ceNeutronData/ceNeutronDatabase_inter.f90
```diff
@@ -196,7 +197,31 @@ subroutine updateMicroXSs(self, E, nucIdx, rand)
     end subroutine updateMicroXSs
 
     !!
-    !! Subroutine to get the elastic scattering majorant cross section in a nuclide
+    !! Subroutine to retrieve the nuclide total majorant cross section
+    !! over a calculated energy range (needed for TMS) and update the nuclide
```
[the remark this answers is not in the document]
valeriaRaffuzzi: > It refers to the span of relative energies a nuclide can see given a material temperature, right?

Yes, exactly. I tried to leave the comment generic because potentially you can use this function with an arbitrary energy range if you wanted too. It doesn't have to be specific to TMS. I can be more specific in the comments though

Tallies/TallyClerks/collisionClerk_class.f90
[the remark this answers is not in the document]
valeriaRaffuzzi: Yeah I agree that it should be made more clear. I tried to keep the code as simple as possible but clearly it's not an easy task and I might have overcomplicated interpretation. 
The fact that colling `getTotalMatXS` is not safe is not great. I'll have a look.

ParticleObjects/particle_class.f90
valeriaRaffuzzi: NO it's not, I think I put it there automatically.

CollisionOperator/CollisionProcessors/neutronCEimp_class.f90
valeriaRaffuzzi: Well if it isn't then I'm happy to revert back and save it in collDat!

valeriaRaffuzzi:
> * QUESTION: should the sampled energy be used for scatterInLab and ScatterFromFixed too????????

@Mikolaj-A-Kowalski @ChasingNeutrons I can see you didn't read the PR description!!! Does anyone have an intuition on this?

Tallies/TallyClerks/collisionClerk_class.f90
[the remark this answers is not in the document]
valeriaRaffuzzi: I don't think separating deterministic vs random xss is a good way to do it. When using URR there is no point in asking for the deterministic xss. Same with TMS, one either wants the temperature majorant or the relative energy xss. What I am thinking for now is: 

```
updateTrackingMatXS:
if TMS -> get Temperature majorant
else   -> get Total

updateTotal: 
if TMS -> get relative energy xss (when doing the relative energy sampling one might as well get the whole package) and return total
else   -> get Total as always

updateMacroXSS:
if TMS -> get relative energy xss
else   -> get Macro as always 
```
So as far as TMS is concerned, updateTrackingMatXS gets the deterministic majorant and updateTotal and updateMacro give the relative energy xss. When TMS is not used, there is no difference between updateTracking and updateTotal.

Mikolaj-A-Kowalski:
> > * QUESTION: should the sampled energy be used for scatterInLab and ScatterFromFixed too????????
> 
> @Mikolaj-A-Kowalski @ChasingNeutrons I can see you didn't read the PR description!!! Does anyone have an intuition on this?

Sorry. It does appear that I have missed this point in the description. The functions should use the LAB frame incident energy, not the relative energy. Basically the sampled energy is 'kind off' an energy in the CoM frame of the collision. If it was used weird things might appear.

valeriaRaffuzzi:
Apologies for adding yet more stuff! I implemented all of the reviews:

- A lot of them were typos, aesthetic stuff and some docs.
- I removed collisionE from the particle. Now the collision energy is stored in collDat. 
- I changed slightly the nuclear data interface (sorry!!). I added a function to the nuclearDatabase: getTrackMatXS. This will be equal to the total material xs normally; when TMS is used, this is the material temperature majorant. The material tracking xs is now stored in ceNeutronCache for each material with the corresponding energy. Now getTotalMatXS and getMacroXSs give consistent results: they give the regular xss OR the relative energy xss in the case of TMS. 

Let me know if this works for you!

Mikolaj-A-Kowalski (approved):
I guess the only thing left to do is integrate and perhaps open an issue about trying to rework cache so it is clearer when we have time/ideas.

PR #122, opened 2024-04-15
1D tally maps

valeriaRaffuzzi:
Just tidying the maps a bit: if we want to keep the separation between tallyMaps and tallyMaps1D, which is useful for the multiMaps, I think some order makes things clearer. 

The main changes are:
- moving all the 1D maps and tests into a new folder: TallyMaps/Maps1D
- renaming sphericalMap -> sphericalRadMap, given that only the radial dimension was mapped. If one wants to make a multidimensional sphericalMap, then that would be a TallyMap
- Making a version of cylindricalMap with only the radial components: /Maps1D/cylindricalRadMap
- Adding a directionMap to map the direction of flight of the particles. For now only a linear grid (0-2pi) is allowed, and the user needs to choose the 2D plane of interest. 

I forgot to update the User manual, I'll do it after the reviews to make sure I am not missing anything given that a lot has to be changed.


Tallies/TallyMaps/Maps1D/directionMap_class.f90
```diff
@@ -0,0 +1,194 @@
...
+
+    call out % startArray(name, [2,self % N])
+
+    do i = 1, self % N
+      ! Print lower bin boundary
+      call out % addValue(self % bounds % bin(i) + PI)
+
+      ! Print upper bin boundar
+      call out % addValue(self % bounds % bin(i + 1) + PI)
+    end do
```
valeriaRaffuzzi, line 173:
```suggestion
    do i = 1, self % N
      ! Print lower bin boundary
      call out % addValue(self % bounds % bin(i))

      ! Print upper bin boundar
      call out % addValue(self % bounds % bin(i + 1))
    end do
```
NOTE: otherwise the bins don't coincide with the bins of the results.

Tallies/TallyMaps/cylindricalMap_class.f90
valeriaRaffuzzi, line 1: Lines 379 - 382: also remove PI from the printed bin values

ChasingNeutrons:
One question before I dive in: would it be neater to have a generic radial map where you can specify which dimensions/plane/axis to account for? For example, it could be xy, xz, yz, or xyz (spherical). This is as opposed to having separate cylindrical and spherical radial maps.

valeriaRaffuzzi:
> One question before I dive in: would it be neater to have a generic radial map where you can specify which dimensions/plane/axis to account for? For example, it could be xy, xz, yz, or xyz (spherical). This is as opposed to having separate cylindrical and spherical radial maps.

Yes that should be easy enough! Let me give it a quick try.

Tallies/TallyMaps/Maps1D/directionMap_class.f90
```diff
@@ -0,0 +1,194 @@
...
+  !!   tallyMap1D interface
+  !!
+  !! NOTE:
+  !!   Behaviour of points exactly at the boundary of bins is undefined.
+  !!   particle can end-up in either of the two
+  !!
+  !! Sample Dictionary Input:
+  !!  directionMap {
+  !!    type directionMap;
+  !!    #plane xz;#            // Optional. Default xy
```
ChasingNeutrons, line 28: For consistency with, say, cylinders and plotting, wouldn't it be preferable to have this as the perpendicular axis? E.g., xy would be z.

Tallies/TallyMaps/Maps1D/directionMap_class.f90
```diff
@@ -0,0 +1,194 @@
...
+      case('xz')
+        self % DIM1 = X_AXIS
+        self % DIM2 = Z_AXIS
+
+      case('xy')
+        self % DIM1 = X_AXIS
+        self % DIM2 = Y_AXIS
+
+      case default
+        call fatalError(Here, 'Keyword orientation must be x, y or z. It is: '//type)
```
ChasingNeutrons, line 84: This or the cases would need to change.

Tallies/TallyMaps/Maps1D/directionMap_class.f90
```diff
@@ -0,0 +1,194 @@
+module directionMap_class
```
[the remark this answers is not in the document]
valeriaRaffuzzi: No worries, I understood! It's a bit of a long name. I personally would leave it as it is until a polarMap comes by. Also because you can map direction on any plane with this, so it's quite general.

Tallies/TallyMaps/Maps1D/directionMap_class.f90
```diff
@@ -0,0 +1,194 @@
...
+        self % DIM2 = Y_AXIS
+
+      case default
+        call fatalError(Here, 'Keyword orientation must be x, y or z. It is: '//type)
+
+    end select
+
+    ! Build grid
+    call dict % get(self % N, 'N')
+    call self % bounds % init(-PI, PI, self % N, 'lin')
```
ChasingNeutrons, line 90: Why from -pi to pi rather than 0 to 2pi? I would say the latter is more common in describing the azimuthal angle.

Tallies/TallyMaps/Maps1D/directionMap_class.f90
valeriaRaffuzzi: I don't know because in a cylinder the perpendicular axis has more of a geometrical meaning than here. Here you're just literally scoping the angle in a 2D plane so I think the plane has more meaning

Tallies/TallyMaps/Maps1D/radialMap_class.f90
```diff
@@ -0,0 +1,267 @@
...
+        allocate(self % axis(2), self % origin(2))
+        self % axis = [X_AXIS, Y_AXIS]
+
+      case('xyz')
+        allocate(self % axis(3), self % origin(3))
+        self % axis = [X_AXIS, Y_AXIS, Z_AXIS]
+        spherical = .true.
+
+      case default
+        call fatalError(Here, 'Keyword orientation must be x, y, z or nothing for spherical. It is: '//type)
```
ChasingNeutrons, line 94: This error is inconsistent with the spherical case above.

Tallies/TallyMaps/Tests/cylindricalMap_test.f90
```diff
@@ -179,7 +179,7 @@ end subroutine testBinNumber
 
   !!
   !! Test correctness of print subroutine
-  !! Does not checks that values are correct, but that calls sequance is without errors
+  !! Does not checks that values are correct, but that calls sequence is without errors
```
ChasingNeutrons, line 182: Does not check* ... but that call* sequence

Tallies/TallyMaps/Maps1D/directionMap_class.f90
valeriaRaffuzzi: Just for convenience, because the Fortran function atan2 returns an angle in the range [-PI, PI].

Tallies/TallyMaps/Maps1D/directionMap_class.f90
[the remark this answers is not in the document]
valeriaRaffuzzi: In this case, it might be even more convenient to leave it generic and add the possibility to modify the angle range to anything within -PI to PI. So one could tally -PI to PI on a plane and 0 to PI on another

Tallies/TallyMaps/Maps1D/directionMap_class.f90
ChasingNeutrons: Fair enough.

Tallies/TallyMaps/Maps1D/directionMap_class.f90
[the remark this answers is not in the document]
ChasingNeutrons: I suppose it would still be a fraction of the azimuthal plane. Wouldn't the polar equivalent need some weighting by a sine? But, in principle, the range could be limited. I imagine that might make things a bit messier and also wouldn't be requested very frequently so I wouldn't worry much about it.

One more thing: which direction would the 0 angle correspond to in a given plane? Or pi/2? It would be good to mention these explicitly at least in the comments.

Mikolaj-A-Kowalski (approved):
Looks good to me with just couple of minor issues. 
Thanks for the addition!

Tallies/TallyMaps/Maps1D/radialMap_class.f90
```diff
@@ -0,0 +1,267 @@
...
+
+    end select
+
+    ! Check & load origin
+    if (spherical) then
+      call dict % getOrDefault(self % origin, 'origin', [ZERO, ZERO, ZERO])
+    else
+      call dict % getOrDefault(self % origin, 'origin', [ZERO, ZERO])
+    end if
```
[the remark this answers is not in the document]
valeriaRaffuzzi: I guess I was imitating what done in cylinderMap, where the origin only had two entries. I am changing both cylinderMap and radialMap to read origins with 3 entries to avoid confusion for the user, but then only the two relevant values are saved. In radialMap, where in the norm2 calculation one wants arbitrarily 2 or 3 values according to whether the map is spherical or cylindrical, having an allocatable origin is more convenient.

Tallies/TallyMaps/Maps1D/directionMap_class.f90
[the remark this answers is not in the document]
valeriaRaffuzzi: I think that if you want to integrate over a polar angle you need to add the sign weighting as a Response, rather than in the tally Map. Now I added a custom min and max angle, so that it generalises better.

Tallies/TallyMaps/Maps1D/directionMap_class.f90
```diff
@@ -0,0 +1,194 @@
...
+  use particle_class,     only : particleState
+  use outputFile_class,   only : outputFile
+  use tallyMap1D_inter,   only : tallyMap1D, kill_super => kill
+
+  implicit none
+  private
+
+  !!
+  !! Maps the particle direction in linear bins in the range 0-360 degrees
+  !!
```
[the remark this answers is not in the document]
valeriaRaffuzzi: Yeah it was. Now I changed it to be defined in degrees from -180 to 180, given that degrees are more intuitive I think. I added those comments.

valeriaRaffuzzi:
I'll ho ahead and merge this, given that it was approved and only a couple of aesthetic things changed since!

Issue #127, opened 2024-06-12
Incorrect units for precursor decay constants


ChasingNeutrons:
ACE files contain the precursor decay constant in [shakes^-1] but we read them (in fissionCE) and state that the value is in units of [s^-1]. We should do a unit conversion of 10^8 shakes/s to obtain the correct values given our chosen units.

ChasingNeutrons:
Resolved by #129

Issue #119, opened 2024-04-05
Check on presence of S(alpha,beta) nuclide


ChasingNeutrons:
I was running a calculation with a moderating nuclide given S(alpha,beta) data. It would run for one iteration and crash mysteriously on the second iteration, reporting neutrons having zero energy.

It turns out that the problem was I had forgotten to include the density of the moderating nuclide! That is, I could specify 'moder {1001.03 <Sab file>;} without actually having 1001.03 present in the material.

Hence I propose adding a check when applying Sab data to ensure that the given nuclide is indeed present in the material.

ChasingNeutrons:
Resolved by #129

Issue #115, opened 2024-02-05
Checks on URR and Sab


valeriaRaffuzzi:
For now there's no check on overlaps between URR and Sab energy ranges. Honestly that should never happen, but you never know with newer (or faulty) nuclear data. 

We should implement a check and give priority to one or the other (I think Sab) if this ever happens.

ChasingNeutrons:
Resolved by #129

PR #129, opened 2024-07-30
S(alpha,beta) stochastic mixing

ChasingNeutrons:
This PR adds stochastic interpolation/mixing to S(alpha,beta).

This is invoked quite simply by adding a second S(a,b) file to the material definition. It required a minor change such that this data is fed (even without mixing, so only one entry) as an array, rather than just as a word as before.

The main thing this required in the code was adding temperature in quite a few places (calls for updating XSs), but this might be something we'll be glad to have in the longer term, e.g., with multiphysics?

Unit and integration tests are passing. I've made some small efforts on the thermalScattering iTest - will add more when I have figured out some weirdness!

This also fixes a few of the issues mentioned, just while I was in there. #127 #119 and #115 should be resolved.


NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
```diff
@@ -912,27 +925,35 @@ subroutine init(self, dict, ptr, silent )
...
-        name = trim(mat % nuclides(j) % toChar())
-
-        if (mat % nuclides(j) % hasSab) then
-          zaid = trim(mat % nuclides(j) % toChar())
-          file = trim(mat % nuclides(j) % file_Sab)
-          name = zaid // '+' // file
-          deallocate(zaid, file)
-        end if
-
+        name = self % makeNuclideName(mat % nuclides(j))
```
valeriaRaffuzzi, line 933: We could move this procedure inside materialMenu, and save the name inside nuclideInfo, so that we don't need to repeat this several times

valeriaRaffuzzi (changes requested):
It all looks good, I just made a few comments! 

The main scary thing is saving flags in the nuclide cache. Let's try to have a think of all the ways it can go wrong!

NuclearData/ceNeutronData/aceDatabase/aceNeutronDatabase_class.f90
ChasingNeutrons: The only problem with this is transferring the Sab discrimination info, i.e., the symbols used to staple things together. They could become universal variables, but feels unnecessary. If there is a case for using it in more than one place though I would be behind it.

valeriaRaffuzzi: I just proposed this because this function is already called twice! No big deal, but maybe it would be neater.

valeriaRaffuzzi (approved):
I think now it all looks good! 

I would just make an issue about the cache, although it's hard to avoid when you're sampling. I think this is part of a bigger issue of rethinking the cache logic a bit.

PR #131, opened 2024-09-02
Update .gitignore to ignore build folder

valeriaRaffuzzi:
As for Esmae's suggestion, this PR is listing the build folder in .gitignore so that it's not tracked, to prevent accidents.


Mikolaj-A-Kowalski (approved):
I mean... the only question is if there is anything that instantly comes to mind as well. 
To me, not.

PR #132, opened 2024-09-10
Changing velocity to speed

ChasingNeutrons:
A small change. I have changed the slightly misleadingly named 'getVelocity' function in the particle to 'getSpeed'. This is because it is the scalar value we are returning, not the vector. I changed this in the particle, in the test, and in the density response. Hopefully I have included all relevant comments too.


valeriaRaffuzzi (approved):
There is no mention of speed in densityResponse_test, right?

PR #134, opened 2024-09-16
Added cells supporting unions

ChasingNeutrons:
There is a new cell type which supports unions, complements, and parentheses. It uses infix notation, inspired by the results of Paul Romano et al.'s recent paper.

There are several style/code structure questions that should be resolved here. 

First, this cell essentially supercedes the simpleCell. Do we want to get rid of the simpleCell? It might be code duplication otherwise and I can't think of a very good case for keeping it. Maybe we wait a little while until we're comfortable with this one.

Second, for the operators necessary, reading them in alongside the surfaces required a deviation in input parsing/reading. Currently lists do not allow mixed number/character arrays. Hence I created the 'tokenArray' to contain arbitrary white-space separated characters of any format. These are read into a character array - I convert the numbers back to integers inside unionCell. It is denoted by using square brackets. Is this acceptable? Is there a nicer way of doing this? This also requires that surfaces and operators be delimited by spaces - maybe this isn't strictly necessary if I put some more logic inside the unionCell itself, for example. I would especially like to resolve this quickly because it will affect the input format and it would be great to settle that quickly so that we can start using this, e.g., with csg2csg.

I have also written a unit test and changed some comments in other nearby files.

### DIFF
--- a/DataStructures/dictParser_func.f90
+++ b/DataStructures/dictParser_func.f90
@@ -44,7 +44,7 @@ module dictParser_func
 contains
 
   !!
-  !! Reads a contents of the file in a stream fashion
+  !! Reads contents of the file in a stream fashion
   !!
   !! All relevant preprocessing is done while reading:
   !!   - TABS are replaced with SPACE
@@ -125,13 +125,15 @@ contains
   !! <DICTIONARY> ::= <ITEM>+                           ! Dictionary consists of 1 or more ITEMS
   !! <ITEM>       ::= <ENTRY>|<KEYWORD>{<DICTIONARY>}
   !! <ENTRY>      ::= <KEYWORD>' '<CONTENT>;            ! At least 1 space in-between <keyword> & <content>
-  !! <CONTENT>    ::= <SINGLE_CONTENT>|(<LIST_CONTENT>)
+  !! <CONTENT>    ::= <SINGLE_CONTENT>|(<LIST_CONTENT>)|(<TOKENARRAY_CONTENT>)
   !! <SINGLE_CONTENT> ::= Int|Real|Word
   !! <LIST_CONTENT> ::= <IntList>|<realList>|<charList>
-  !! <IntList> ::= Space Separated Integers e.g.: 1 2   3
-  !! <realList> ::= Space separated Numbers e.g.: 1.0 2   3.0E+2
-  !! <charList> ::= Space separated not-numbers e.g.: 1.0\ word       char
-  !! <KEYWORD>  ::= Character string with len_trim < nameLen constant
+  !! <TOKENARRAY_CONTENT> ::= <tokenArray>
+  !! <IntList>    ::= Space Separated Integers e.g.: 1 2   3
+  !! <realList>   ::= Space separated Numbers e.g.: 1.0 2   3.0E+2
+  !! <charList>   ::= Space separated not-numbers e.g.: 1.0\ word       char
+  !! <tokenArray> ::= Space separated anything, converted to chars e.g.: 3  -eight <    () 
+  !! <KEYWORD>    ::= Character string with len_trim < nameLen constant
   !!
   !! Important facts to note:
   !!  1) Keywords are Case Sensitive keYwORD /= keyword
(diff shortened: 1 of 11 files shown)


Geometry/Cells/Tests/unionCell_test.f90
```diff
@@ -0,0 +1,166 @@
...
+  & surf2 { id 15; type box; origin (0.0 0.0 0.0); halfwidth ( 4.6 4.6 4.6);} &
+  & surf3 { id 4;  type xTruncCylinder; origin (0 0 0); halfwidth 5.1; radius 1.0;} &
+  & surf4 { id 5;  type yTruncCylinder; origin (0 0 0); halfwidth 5.1; radius 1.0;} &
+  & surf5 { id 6;  type zTruncCylinder; origin (0 0 0); halfwidth 5.1; radius 1.0;} "
+
+  ! Note that fill is not really needed to build a cell. It is used by cellShelf only
+  character(*), parameter :: EASYCELL_DEF = "&
+  & id 2; type unionCell; surfaces [-13 -15 ]; filltype outside; "
+  character(*), parameter :: COMPLEXCELL_DEF = "&
+  & id 3; type simpleCell; surfaces [< -13 -15 > # < -4 : -5 : -6 > ]; filltype outside; "
```
valeriaRaffuzzi, line 24: shouldn't this be ``unionCell`` too?

Geometry/Cells/Tests/unionCell_test.f90
```diff
@@ -0,0 +1,166 @@
...
+    type(dictionary) :: dict
+
+    call charToDict(dict, SURF_DEF)
+    call surfs % init(dict)
+    call dict % kill()
+    call charToDict(dict, EASYCELL_DEF)
+    call easyCell % init(dict, surfs)
+    call dict % kill()
+    call charToDict(dict, COMPLEXCELL_DEF)
+    call complexCell % init(dict, surfs)
```
valeriaRaffuzzi, line 47:
```suggestion
    call charToDict(dict, SURF_DEF)
    call surfs % init(dict)
    call dict % kill()
    
    call charToDict(dict, EASYCELL_DEF)
    call easyCell % init(dict, surfs)
    call dict % kill()
    
    call charToDict(dict, COMPLEXCELL_DEF)
    call complexCell % init(dict, surfs)
```
What about adding spaces to make it more legible?

Geometry/Cells/Tests/unionCell_test.f90
valeriaRaffuzzi: The cell % init procedure doesn't read ``type``, so that could be removed actually.

Geometry/Cells/unionCell_class.f90
```diff
@@ -0,0 +1,477 @@
...
+  ! Define CSG operators
+  integer(shortInt), parameter :: union     = huge(shortInt) - 1, &
+                                  intersect = union - 1, &
+                                  openBra   = union - 2, &
+                                  closeBra  = union - 3, &
+                                  complem   = union - 4
+
+
+  !!
+  !! CSG cell defined by intersections and unions of halfspaces
```
valeriaRaffuzzi, line 23: And complements?

valeriaRaffuzzi (approved):
Thanks a lot for the development, it's very exciting to finally have unions! 
Just a couple of comments but then it can be merged for me.

valeriaRaffuzzi:
To answer the questions in the description:

- if the new union cell includes all the functionalities of the simple cell and it's not any slower, then I agree there is no reason to keep the simpleCell
- After reviewing the code I thought the tokenArray parser was neat enough. I see why it's convenient to read everything as characters!

DataStructures/dictParser_func.f90
```diff
@@ -437,7 +443,39 @@ subroutine readEntry(dict, start, end, tape)
         end if
       end if
 
+    elseif (l3 == 0 .neqv. l4 == 0) then ! Only single bracket is present
```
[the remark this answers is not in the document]
ChasingNeutrons: I believe that should be there already a few lines up. This code is basically a duplicate of that for l1 and l2, just for a different bracket type.

Geometry/Cells/Tests/unionCell_test.f90
ChasingNeutrons: Very good spot. I will remove it from the equivalent simpleCell test for now too.

ChasingNeutrons:
Thanks for the comments.

I think, at least for this PR, I will keep simple cells and open an issue to remove them. In part this is because getting rid of it means changing all of the inputs which use them!

valeriaRaffuzzi:
> Thanks for the comments.
> 
> I think, at least for this PR, I will keep simple cells and open an issue to remove them. In part this is because getting rid of it means changing all of the inputs which use them!

That's ok, we can make an issue. Merge?

PR #135, opened 2024-09-19
Quadric surface

ChasingNeutrons:
Adds a general quadratic shape of the form:
F(x,y,z) = Ax^2 + By^2 + Cz^2 + Dxy + Eyz + Fxz + Gx + Hy + Jz + I
Also removed dotProduct from surfaces in favour of intrinsic dot_product.


valeriaRaffuzzi (changes requested):
All good to me! The only doubt is about the surface tolerance, I am not convinced about that but I don't think I have a better suggestion for now. We should think about this a bit more I think.

Issue #31, opened 2023-03-16
Installation Mac 


aliciadrm:
Hi I am trying to install SCONE on my macOS 13 and when running the automated tests I have failures in the plane_test_suite.testHalfspace and in the plane_test_suite.testDistance.
My compiler is GNU Fortran (GCC) 12.2.0.

If you have any insights or advice on how to proceed, please let me know.

Thanks

ChasingNeutrons:
Hi @aliciadrm,

I have seen something like this before when using -march=native in the compiler options. Is the error similar for you: there's a failure in a distance calculation? It should be returning a very large number which isn't quite equal to the INFINITY value we set in the SharedModules/universalVariables.f90 file. If so, you can proceed relatively safely for now. While we should hunt down what's going on, one thing may be to try a slightly older gfortran version. I am usually using 9.4.0, but I am sure we have successfully past all tests with more recent versions too.

Paul

thltsui:
Hi, I am trying to install SCONE on an Apple M2 Max (M2 chip) with macOS 13.5.2. 

When I first tried to install according to step 5 under Compiling SCONE under https://scone.readthedocs.io/en/latest/Installation.html,
I received an error message that tells me to turn LTO off.
Then, I tried to run
cmake -E chdir ./Build cmake ./.. -DLTO=OFF,
but again, the installation got stuck.
I re-ran the code again and got the following messages:

-- Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES) 
-- Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES) 
-- Could NOT find OpenMP (missing: OpenMP_C_FOUND OpenMP_CXX_FOUND) (found version "4.5")
-- /Library/Developer/CommandLineTools/SDKs/MacOSX13.3.sdk/System/Library/Frameworks/Accelerate.framework-lm-ldl
-- 
--  ><((((*>  ><((((*>  ><((((*>  ><((((*>  ><((((*>  ><((((*> 
--   SCONE CONFIGURATION SUMMARY
--   Unit Tests:                 OFF
--   Link-time optimisation:     OFF
--   Code coverage logging:      OFF
--   Extra runtime debug checks: OFF
--   OpenMP parallelism:         ON
--   Fortran compiler:     /opt/homebrew/bin/gfortran
--   Compiler version:     13.2.0
[...]

PR #139, opened 2024-10-24
macOS support

NRavoisin96:
[SCONE Installation guide for macOS.docx](https://github.com/user-attachments/files/17506021/SCONE.Installation.guide.for.macOS.docx)

Dear all,

After longer than I would like to admit, I am happy to announce that SCONE now runs natively on macOS, both for Intel and ARM (Apple Silicon) machines. I have attached the documentation to correctly set up SCONE on macOS. A couple points to note:

- Most of the issues reported by students seem to stem from improper configuration and / or compiler choices. Apple devices, by default, ship with Apple's in-house C compiler, known as Apple Clang. It is _strongly_ recommended to use this compiler when installing pFUnit and SCONE, as other compilers such as the ones in gcc packages are not officially supported by Apple and will likely cause errors when trying to set up SCONE. As such, in the documentation it is _required_ to use Apple Clang.
- A special mention should be made regarding issue #31: the reason for the unit tests failing in this case is due to rounding-off errors being slightly different on ARM-based Macs compared to machines (including Macs) running on Intel CPUs. This can be due to a combination of several factors, including different memory registry lengths and / or instruction set optimisations. In any case, the actual culprit causing plane test to fail is given here:
 `k = dotProduct(u, self % norm)`
`c = self % evaluate(r)`
`if ( k == ZERO .or. abs(c) < self % surfTol()) then ! Parallel or at the surface`
`...`
Here, k (a floating point number) is tested to be equal to ZERO (another floating point number) _to the bit_. This is bad practice and dangerous. I have written two small helper functions in the `genericProcedures.f90` module which test for equality between two floating point numbers using small tolerances instead. In the future, please make use of these instead of testing for perfect equality, unless you are 100% sure that the variable you are testing _can_ be exactly equal to the specific value of interest. In general, any variable which results from mathematical operations should never be tested for perfect equality. I have corrected instance of such tests in the Geometry section, although I have seen it happen in many other sections. As I am less familiar with the code in these sections, I would advise people working on them to have a look and assess when exact equality can be tolerated and when it needs to be replaced with the helper functions.
- Fortran comes with an intrinsic `dot_product` function so there is no need to define our own `dotProduct` function in the `genericProcedures.f90` module. I have removed it in this pull request.

Many thanks,
Nathan

### DIFF
--- a/docs/Installation.rst
+++ b/docs/Installation.rst
@@ -129,7 +129,7 @@ needs to be available::
 
 Alternatively, if one does't have access to mpi_f08 and wants to compile SCONE without MPI support::
 
-     cmake -DSKPI_MPI=YES ./..
+     cmake -DSKIP_MPI=YES ./..
 
 #. Run make and install application
 


SharedModules/genericProcedures.f90
```diff
@@ -594,6 +598,88 @@ function concatenateArrays_Real(array1,array2) result(out)
...
+    ! If reached here, value and target are not within absolute or relative tolerance of each 
+    ! other. Update equal = .false.
+    equal = .false.
+
+  end function areEqual_defReal
+
+  !!
+  !! Returns .true. if all elements of an array are equal to target. 
+  !!
+  !! Due to floating point artihmetic and rounding-off errors being slightly different
```
ChasingNeutrons, line 641:
```suggestion
  !! Due to floating point arithmetic and rounding-off errors being slightly different 
```

SharedModules/genericProcedures.f90
```diff
@@ -594,6 +598,88 @@ function concatenateArrays_Real(array1,array2) result(out)
 
   end function concatenateArrays_Real
 
+  !!
+  !! Returns .true. if value and target are equal. 
+  !!
+  !! Due to floating point artihmetic and rounding-off errors being slightly different
```
ChasingNeutrons, line 604:
```suggestion
  !! Due to floating point arithmetic and rounding-off errors being slightly different 
```

ChasingNeutrons:
Thanks for doing this @NRavoisin96, it has been long overdue. I came across that dotProduct recently and wondered why we didn't use dot_product - probably because it is some quite old code and we didn't know any better! Likewise, thanks for the 'areEqual', that has also bugged me in the past.

The code updates look great. The only thing is to update the docs. That is, the contents of your guide should be incorporated into docs/Installation.rst, rather than as a separate document which may get lost on closing this PR.

NRavoisin96:
@ChasingNeutrons, 

The documentation has been updated. I have also used this opportunity to fix some small typos in the installation.rst file and added a section at the on how to actually run SCONE (using the JEZ input file as an example), so that the required syntax to execute SCONE is actually included somewhere in the documentation.

Many thanks,
Nathan

docs/Installation.rst
```diff
@@ -114,147 +116,271 @@ This is only required if the unit tests are to be build.
...
+	/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+
+#. Once `Homebrew` is installed, type the following command in your `Terminal` window. 
+   This will install the latest versions of all the packages required to correctly set 
+   up and run SCONE::
+
+     brew install gcc cmake python git openblas lapack libomp
+
+#. Close your `Terminal` window. Open a new `Finder` window and navigate to your `Home` directory 
+   (⌘ + ⇧ + h). Display hidden files (⌘ + ⇧ + .) and find the ``.zprofile`` file (this is the macOS
```
Mikolaj-A-Kowalski, line 194: :heart: unicode!

docs/Installation.rst
```diff
@@ -114,147 +116,271 @@ This is only required if the unit tests are to be build.
...
+
+     brew install gcc cmake python git openblas lapack libomp
+
+#. Close your `Terminal` window. Open a new `Finder` window and navigate to your `Home` directory 
+   (⌘ + ⇧ + h). Display hidden files (⌘ + ⇧ + .) and find the ``.zprofile`` file (this is the macOS 
+   equivalent of the ``.bashrc`` file on Linux distributions). Open it and insert **any of the 
+   following lines which are not already present** (note: this depends on whether you have a Mac 
+   running on an Intel CPU or an ARM - Apple Silicon - chip):
+   
+   * Intel::
```
Mikolaj-A-Kowalski, line 199: I don't quite follow why there are two sections based on the CPU ISA?
The only difference seem to be the install path of the `homebrew` package manager. Maybe we can combine both sections (and just point out the difference at the start) to minimise repetition?

docs/Installation.rst
```diff
@@ -114,147 +116,271 @@ This is only required if the unit tests are to be build.
...
+	git clone https://github.com/Goddard-Fortran-Ecosystem/pFUnit.git
+	cd pFUnit
+	mkdir build
+	cd build
+
+#. Before proceeding, **make sure that the default C compiler is Apple Clang by entering 
+   the following command**::
+
+     gcc --version
+
```
Mikolaj-A-Kowalski, line 261: I don't follow this part. If we verify `clang` why do we want to invoke `gcc`? Is this a typo?

docs/Installation.rst
```diff
@@ -114,147 +116,271 @@ This is only required if the unit tests are to be build.
...
+    
+     to ensure that the default C compiler is Apple Clang. Now initialise CMake (you should 
+     still be in the build folder on your `Terminal`) by typing::
+
+	cmake ./..
+
+   * Initialise CMake by specifying which C compiler to use. In your `Terminal` window enter 
+     the following::
+
+	cmake -D CMAKE_C_COMPILER=CLANG ./..
```
Mikolaj-A-Kowalski, line 280: Are you sure it  needs to be capitalised? I though it was supposed to be a path or command that invokes the compiler. Hence I was expecting `clang`...

Also I am not sure why are the steps with `gcc` above necessary. Sorry, I got a bit lost here.

docs/Installation.rst
```diff
@@ -263,23 +389,54 @@ its own library file. An example of it is given in *IntegrationTestFiles/testLib
...
+
+          ...
+
+You may replace the ``$SCONE_ACE`` environmental variable in the `JEZ` file with the absolute path to the ``fullLib.xsfile`` 
+file or, better yet, export this variable in your ``.bashrc``/``.zprofile`` file (depending on your OS) by adding the 
+following line::
+     
+     export SCONE_ACE=path_to_fullLib.xsfile
+
+and saving the changes in either case. Note that if you export this variable in your ``.bashrc``/``.zprofile`` file, you
```
Mikolaj-A-Kowalski, line 436: I wouldn't insist on setting the environmental variables in the `.basrc`. I think it is fine to export them before running. It makes (should make) one more conscious of the nuclear data they use ;-)

Mikolaj-A-Kowalski (commented):
Sorry it took me so long to get around to this. 
In addition to the minor comments/questions I have one major issue. I don't think it is a good idea to make the surfaces even fuzzier due to the one unit test failing. Especially that it fails mostly because it is too strict. The distance calculated on M# chips is still arguably correct (is supposed to be very large, and it is). I think we can just allow for more leeway in the test to solve our problem. My fear is that reasoning about edge cases in surfaces is quite tricky as it is. Making the comparisons not exact can make that problem worse. 

That being said the generic function for FP equality is welcome and should be integrated.

PR #141, opened 2024-11-15
MPI

valeriaRaffuzzi:
Finally! I think this is ready to go. 

This PR includes MPI support. The main differences are in the tallies and in the dungeon for population normalisation between cycles and load balancing (from Paul Romano's PhD). 

In the tallies, one key change is that a report was added: closeCycle other than reportCycleEnd. With MPI, there are two options: mpiSync 1 means that the tallies are synchronised each cycle; mpiSync 0 means that the tallies from each rank are collected at the end of the calculation. All calculations give identical results when mpiSync 1; they are within statistics when mpiSync 0. NOTE that this option applies to all tallies included in a tally admin rather to individual clerks. Splitting reportCycleEnd into two procedures (i.e., adding closeCycle), makes reproducibility easier for most tallyClerks.

In the dungeon, population normalisation was implemented using a new data structure, heapQueue. Note that to ensure reproducibility, particles have to be sorted before and after sampling. Then, load balancing is performed transferring particles between processes.

Results seem to be reproducible for all tallies, and all the tests pass successfully.


PhysicsPackages/eigenPhysicsPackage_class.f90
```diff
@@ -3,10 +3,18 @@ module eigenPhysicsPackage_class
...
   use universalVariables
   use endfConstants
-  use genericProcedures,              only : fatalError, printFishLineR, numToChar, rotateVector
+  use genericProcedures,              only : numToChar, rotateVector
+  use display_func,                   only : printFishLineR, statusMsg, printSectionStart, &
+                                             printSectionEnd, printSeparatorLine
+  use mpi_func,                       only : isMPIMaster, getWorkshare, getOffset, getMPIRank
+#ifdef MPI
+  use mpi_func,                       only : MASTER_RANK, MPI_Bcast, MPI_INT, MPI_COMM_WORLD, &
+                                             MPI_DOUBLE, mpi_reduce, MPI_SUM
```
ChasingNeutrons, line 12:
```suggestion
                                             MPI_DOUBLE, MPI_REDUCE, MPI_SUM
```

PhysicsPackages/eigenPhysicsPackage_class.f90
```diff
@@ -277,22 +308,29 @@ subroutine cycles(self, tally, tallyAtch, N_cycles)
       end_T = real(N_cycles,defReal) * elapsed_T / i
       T_toEnd = max(ZERO, end_T - elapsed_T)
 
+#ifdef MPI
+      ! Print the population numbers referred to all processes to screen
+      call mpi_reduce(nStart, nTemp, 1, MPI_INT, MPI_SUM, MASTER_RANK, MPI_COMM_WORLD, error)
```
ChasingNeutrons, line 313:
```suggestion
      call MPI_REDUCE(nStart, nTemp, 1, MPI_INT, MPI_SUM, MASTER_RANK, MPI_COMM_WORLD, error)
```

PhysicsPackages/eigenPhysicsPackage_class.f90
```diff
@@ -3,10 +3,18 @@ module eigenPhysicsPackage_class
   use numPrecision
   use universalVariables
   use endfConstants
-  use genericProcedures,              only : fatalError, printFishLineR, numToChar, rotateVector
+  use genericProcedures,              only : numToChar, rotateVector
+  use display_func,                   only : printFishLineR, statusMsg, printSectionStart, &
+                                             printSectionEnd, printSeparatorLine
+  use mpi_func,                       only : isMPIMaster, getWorkshare, getOffset, getMPIRank
+#ifdef MPI
+  use mpi_func,                       only : MASTER_RANK, MPI_Bcast, MPI_INT, MPI_COMM_WORLD, &
```
ChasingNeutrons, line 11: Are we consistent with how we write MPI commands? Should they all be all caps or not? If they should then we should have 
```suggestion
  use mpi_func,                       only : MASTER_RANK, MPI_BCAST, MPI_INT, MPI_COMM_WORLD, &
```

PhysicsPackages/eigenPhysicsPackage_class.f90
```diff
@@ -277,22 +308,29 @@ subroutine cycles(self, tally, tallyAtch, N_cycles)
       end_T = real(N_cycles,defReal) * elapsed_T / i
       T_toEnd = max(ZERO, end_T - elapsed_T)
 
+#ifdef MPI
+      ! Print the population numbers referred to all processes to screen
+      call mpi_reduce(nStart, nTemp, 1, MPI_INT, MPI_SUM, MASTER_RANK, MPI_COMM_WORLD, error)
+      nStart = nTemp
+      call mpi_reduce(nEnd, nTemp, 1, MPI_INT, MPI_SUM, MASTER_RANK, MPI_COMM_WORLD, error)
```
ChasingNeutrons, line 315:
```suggestion
      call MPI_REDUCE(nEnd, nTemp, 1, MPI_INT, MPI_SUM, MASTER_RANK, MPI_COMM_WORLD, error)
```

PhysicsPackages/eigenPhysicsPackage_class.f90
```diff
@@ -262,6 +288,11 @@ subroutine cycles(self, tally, tallyAtch, N_cycles)
 
       end select
 
+#ifdef MPI
+      ! Broadcast k_eff obtained in the master to all processes
+      call MPI_Bcast(k_new, 1, MPI_DOUBLE, MASTER_RANK, MPI_COMM_WORLD)
```
ChasingNeutrons, line 293:
```suggestion
      call MPI_BCAST(k_new, 1, MPI_DOUBLE, MASTER_RANK, MPI_COMM_WORLD)
```

PhysicsPackages/eigenPhysicsPackage_class.f90
```diff
@@ -429,6 +471,12 @@ subroutine init(self, dict)
       call FNV_1(string,seed_temp)
 
     end if
+
+    ! Broadcast seed to all processes
+#ifdef MPI
+    call MPI_Bcast(seed_temp, 1, MPI_INT, MASTER_RANK, MPI_COMM_WORLD)
```
ChasingNeutrons, line 477:
```suggestion
    call MPI_BCAST(seed_temp, 1, MPI_INT, MASTER_RANK, MPI_COMM_WORLD)
```

PhysicsPackages/fixedSourcePhysicsPackage_class.f90
```diff
@@ -3,10 +3,17 @@ module fixedSourcePhysicsPackage_class
   use numPrecision
   use universalVariables
   use endfConstants
-  use genericProcedures,              only : fatalError, printFishLineR, numToChar, rotateVector
+  use genericProcedures,              only : numToChar, rotateVector
+  use display_func,                   only : printFishLineR, statusMsg, printSectionStart, printSectionEnd, &
+                                             printSeparatorLine
+  use mpi_func,                       only : isMPIMaster, getWorkshare, getOffset, getMPIRank
+#ifdef MPI
+  use mpi_func,                       only : MASTER_RANK, MPI_Bcast, MPI_INT, MPI_COMM_WORLD
```
ChasingNeutrons, line 11:
```suggestion
  use mpi_func,                       only : MASTER_RANK, MPI_BCAST, MPI_INT, MPI_COMM_WORLD
```

PhysicsPackages/fixedSourcePhysicsPackage_class.f90
```diff
@@ -347,16 +370,22 @@ subroutine init(self, dict)
...
       call date_and_time(date, time)
       string = date // time
-      call FNV_1(string,seed_temp)
+      call FNV_1(string, seed_temp)
 
     end if
+
+    ! Broadcast seed to all processes
+#ifdef MPI
+    call MPI_Bcast(seed_temp, 1, MPI_INT, MASTER_RANK, MPI_COMM_WORLD)
```
ChasingNeutrons, line 386:
```suggestion
    call MPI_BCAST(seed_temp, 1, MPI_INT, MASTER_RANK, MPI_COMM_WORLD)
```

SharedModules/errors_mod.f90
```diff
@@ -9,6 +9,7 @@ module errors_mod
 
   use numPrecision
   use universalVariables, only : MAX_COL
+  !use mpi_func,           only : getMPIRank
```
ChasingNeutrons, line 12: Delete this?

SharedModules/errors_mod.f90
```diff
@@ -36,7 +37,9 @@ subroutine fatalError(where, why)
 
     ! Upper frame
     write(error_unit, *) repeat('<>', MAX_COL / 2)
-
+! #ifdef MPI
```
ChasingNeutrons, line 40: Delete?

SharedModules/mpi_func.f90
```diff
@@ -0,0 +1,198 @@
...
+  integer(shortInt), private   :: worldSize = 1
+  integer(shortInt), private   :: rank = 0
+  integer(shortInt), parameter :: MASTER_RANK = 0
+
+  !! Public type that replicates exactly particleState
+  !!
+  !! It is necessary for load balancing in the dungeon: particles have to be
+  !! transferred betwen processes, and MPI doesn't allow to transfer types with
+  !! type-bound procedures
+  type, public :: particleStateDummy
```
ChasingNeutrons, line 19: This is so painful :( Are there no alternatives around this? It feels like a nasty bug waiting to happen!

SharedModules/mpi_func.f90
```diff
@@ -0,0 +1,198 @@
...
+  !!
+  !! Initialise MPI environment
+  !!
+  !! Needs to be called at the beginning of calculation before any MPI calls
+  !!
+  subroutine mpiInit()
+#ifdef MPI
+    integer(shortInt)        :: ierr
+
+    call mpi_init(ierr)
```
ChasingNeutrons, line 51: As before, I think we need to be consistent with how we call MPI functions. Maybe I have missed something.

docs/Running.rst
```diff
@@ -0,0 +1,37 @@
...
+Note that instructions on how to write an input file can be found
+in the :ref:`Input Manual<input-manual>`.
+
+Running SCONE
+'''''''''''''
+
+After installation of all the dependencies and the compilation of SCONE, 
+one can finally run the code. The executable ``scone.out`` is in the ``build``
+folder; one can run it from any folder by specifying the path to the executables
+and to the input file::
```
ChasingNeutrons, line 15: Perhaps it would be good to clearly describe the interaction between OpenMP and MPI?

ChasingNeutrons:
To get the tests passing, I think it's a case of including MPI installation in the docker image? mikolajkowalski/scone-test

SharedModules/mpi_func.f90
valeriaRaffuzzi: I don't think so. We need to pass particle states between ranks. The alternative is to pass individual properties but I don't see how that's less painful.

SharedModules/mpi_func.f90
valeriaRaffuzzi: We should. I am not sure if @Mikolaj-A-Kowalski had a plan when starting MPI. We could use full caps for the in-build MPI functions, like MPI_REDUCE and MPI_BCAST, and normal SCONE capitalisation for the SCONE procedures inside mpi_func.

valeriaRaffuzzi:
After a long time spent fighting with the compiler..... he is still winning. The problem is related to the Docker container though, rather than to SCONE. OMPI doesn't like it when we try to run the tests as root-user in the container (this refers to the unitTests). I managed to force it with gfortran9 and gfortran10 (which, as you can see, run ok). However I couldn't find a way to get gfortran8 to agree to run!
The latest is that I commented out gfortran8 to make sure that everything else was working correctly, which it is. If we want to reinstate gfortran8, I welcome ideas!

valeriaRaffuzzi:
Another thing to notice is that in the particleDungeon, the previous sampling w/o replacement algorithm is in a new procedure called samplingWithoutReplacement, and it isn't used. 
I didn't want to delete it completely because it's nice code, and it might be useful in other contexts in the future? If you both agree I can however delete it since now it is redundant.

Mikolaj-A-Kowalski:
> After a long time spent fighting with the compiler..... he is still winning. The problem is related to the Docker container though, rather than to SCONE. OMPI doesn't like it when we try to run the tests as root-user in the container (this refers to the unitTests). I managed to force it with gfortran9 and gfortran10 (which, as you can see, run ok). However I couldn't find a way to get gfortran8 to agree to run! The latest is that I commented out gfortran8 to make sure that everything else was working correctly, which it is. If we want to reinstate gfortran8, I welcome ideas!

The reason here is most probably that the `gfotran-8` image is based on the older Debian (buster), which has version of OpenMPI ([`3.1`](https://packages.debian.org/buster/openmpi-bin)) before [they added](https://github.com/open-mpi/ompi/pull/5598) the environmental variables option in 4.0. 

In general we should probably push up the compiler versions for CI and maybe drop gfortran-8 :-/ 
I would prefer it done in a separate PR for clarity. And in the same PR the newer compilers should be added. 

If we want to keep gfotran-8 we could just try MPICH?

SharedModules/mpi_func.f90
Mikolaj-A-Kowalski: Yep... massive delay here. 
I am generally fully in the team `small_letters` with regards to all function calls. Capitals should just indicate constants etc. 
I would just argue that the ALL CAPS is a bit of a legacy thing from the 80s. We should just treat MPI procedures like we do the intrinsic procedures and keep them small.

valeriaRaffuzzi:
> The reason here is most probably that the `gfotran-8` image is based on the older Debian (buster), which has version of OpenMPI ([`3.1`](https://packages.debian.org/buster/openmpi-bin)) before [they added](https://github.com/open-mpi/ompi/pull/5598) the environmental variables option in 4.0.
> 
> In general we should probably push up the compiler versions for CI and maybe drop gfortran-8 :-/ I would prefer it done in a separate PR for clarity. And in the same PR the newer compilers should be added.
> 
> If we want to keep gfotran-8 we could just try MPICH?

Yes that makes sense, about the older OMPI version (annoying!). I agree about dropping gfortran 8 and adding newer versions (all this in a separate PR). But in this case, the problem remains that I can't get the tests to pass in this PR.... I will surely manage with more work but I wonder if it's worth spending time on this. 
Could try MPICH, but just to then drop it in the next PR?

SharedModules/mpi_func.f90
valeriaRaffuzzi: I think all the procedures should now be in small letters. The types however are all caps, as well as MPI inbuilt stuff like ``MPI_COMM_WORLD``. I could make it all small if we prefer, that's not a problem.

Mikolaj-A-Kowalski:
> > The reason here is most probably that the `gfotran-8` image is based on the older Debian (buster), which has version of OpenMPI ([`3.1`](https://packages.debian.org/buster/openmpi-bin)) before [they added](https://github.com/open-mpi/ompi/pull/5598) the environmental variables option in 4.0.
> > In general we should probably push up the compiler versions for CI and maybe drop gfortran-8 :-/ I would prefer it done in a separate PR for clarity. And in the same PR the newer compilers should be added.
> > If we want to keep gfotran-8 we could just try MPICH?
> 
> Yes that makes sense, about the older OMPI version (annoying!). I agree about dropping gfortran 8 and adding newer versions (all this in a separate PR). But in this case, the problem remains that I can't get the tests to pass in this PR.... I will surely manage with more work but I wonder if it's worth spending time on this. Could try MPICH, but just to then drop it in the next PR?

We can just make the new PR quick (yes... I know) and then rebase this one on the `main` with new compilers when it is merged. This one can be left without gfotran-8 for now.

ParticleObjects/Tests/particleDungeon_test.f90
```diff
@@ -245,8 +262,16 @@ subroutine testNormPopUp()
     type(particleDungeon)    :: dungeon
     type(particle)           :: p
     type(RNG)                :: pRNG
-    integer(shortInt)        :: i
+    integer(shortInt)        :: i, worldSize, ierr
```
ChasingNeutrons: Do these need to be initialised to avoid warnings?

ParticleObjects/particleDungeon_class.f90
```diff
@@ -403,11 +417,292 @@ subroutine normWeight(self,totWgt)
...
+    ! Perform load balancing by redistributing particles across processes
+    if (nRanks > 1) call self % loadBalancing(totPop, nRanks, rank, popSizes)
+#endif
+
+  end subroutine normSize_Repr
+
+  !!
+  !! Perform nearest neighbor load balancing
+  !!
+#ifdef MPI
```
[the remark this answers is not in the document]
valeriaRaffuzzi: I changed it! I see arguments to do it either way but I don't have a strong preference.

PR #142, opened 2024-12-18
Update compiler versions in Docker container

valeriaRaffuzzi:
Upgrade compiler versions consistently with the upgrade in the Docker container.

NOTE: The tests _should_ fail until the PR to upgrade the Docker container is accepted.

### DIFF
--- a/NuclearData/mgNeutronData/baseMgNeutron/Tests/baseMgNeutronDatabase_iTest.f90
+++ b/NuclearData/mgNeutronData/baseMgNeutron/Tests/baseMgNeutronDatabase_iTest.f90
@@ -7,6 +7,7 @@ module baseMgNeutronDatabase_iTest
   use dictionary_class,   only : dictionary
   use dictParser_func,    only : charToDict
   use particle_class,     only : particle
+  use RNG_class,          only : RNG
 
   ! Nuclear Data Objects & Interfaces
   use baseMgNeutronDatabase_class, only : baseMgNeutronDatabase, baseMgNeutronDatabase_CptrCast, &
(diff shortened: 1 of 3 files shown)


Mikolaj-A-Kowalski:
Actually the containers are already there. But we have this single failure in the `debug` job. 
We need to figure out what is going on there with gfortran12.

NuclearData/mgNeutronData/baseMgNeutron/Tests/baseMgNeutronDatabase_iTest.f90
```diff
@@ -207,6 +210,8 @@ subroutine testBaseMgNeutronDatabaseWithP1()
     @assertEqual(4, database % nGroups())
 
     ! Test getting Transport XS
+    ! Associate pointer to pass tests in debug mode
+    p % pRNG => pRNG
```
Mikolaj-A-Kowalski, line 214: Do we know why it was unhappy?
Was it because we provide the pointer later to a function with a non-pointer dummy argument (if it is the case thank you GCC!)

Mikolaj-A-Kowalski (approved):

NuclearData/mgNeutronData/baseMgNeutron/Tests/baseMgNeutronDatabase_iTest.f90
valeriaRaffuzzi: I don't think so. The error I was getting was because the particle gets passed to a few functions (something like ``mgDatabase % getTrackingXS`` -> ``mgDatabase % getTotalXS`` -> ``mgMaterial % getTotalXS``) where p % pRNG is an argument, and the problem was that this is a pointer and it was unassigned.

Issue #145, opened 2025-01-22
Accounting for multiplicative scatter in MG implicit keff estimator


ChasingNeutrons:
I don't think we are currently accounting for multiplicative scattering reactions in MG mode when estimating keff (this does seem to work for CE).

On line 228 of keffImplicitClerk, we check for MT = macroAllScatter. However, in MG, scattering falls under MT = macroIEScatter. Hence, we miss out on the multiplication that can result.

I propose adding macroIEScatter into that same line 228. On a quick test, that seems to give an identical keff to the CE case for an infinite homogeneous problem, using the resulting mg XSs from the mgXsClerk tally.

ChasingNeutrons:
Resolved by #146

PR #146, opened 2025-01-22
Fix to implicit keff for MG scattering

ChasingNeutrons:
This addresses issue #145. Also fixes a few typos around keffImplicitClerk_class


valeriaRaffuzzi (approved):
Looks good, so weird we missed this before! Thanks for this

PR #148, opened 2025-01-22
Small fixes in cones and trackingHT

valeriaRaffuzzi:
Two small fixes:

1. Cones throw a fatal error if the two bases (which define the orientation of the cone) have different signs. This was the case even when one the bases was zero. Now, one of the two is allowed to be zero.
2. HT would throw a fatal error if the tracking method selection happened on a particle in void (because this means a material XS lookup in void was attempted). Now ST is chosen if the particle is in void.

### DIFF
--- a/TransportOperator/transportOperatorHT_class.f90
+++ b/TransportOperator/transportOperatorHT_class.f90
@@ -57,11 +57,15 @@ contains
     ! Get majornat XS inverse: 1/Sigma_majorant
     majorant_inv = ONE / self % xsData % getTrackingXS(p, p % matIdx(), MAJORANT_XS)
 
-    ! Obtain the local cross-section
-    sigmaT = self % xsData % getTrackMatXS(p, p % matIdx())
+    ! Obtain the local cross-section. Always choose ST in void
+    if (p % matIdx() == VOID_MAT) then
+      sigmaT = ZERO
+    else
+      sigmaT = self % xsData % getTrackMatXS(p, p % matIdx())
+    end if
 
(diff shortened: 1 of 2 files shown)


TransportOperator/transportOperatorHT_class.f90
```diff
@@ -57,11 +57,15 @@ subroutine tracking_selection(self, p, tally, thisCycle, nextCycle)
...
     majorant_inv = ONE / self % xsData % getTrackingXS(p, p % matIdx(), MAJORANT_XS)
 
-    ! Obtain the local cross-section
-    sigmaT = self % xsData % getTrackMatXS(p, p % matIdx())
+    ! Obtain the local cross-section. Always choose ST in void
+    if (p % matIdx() == VOID_MAT) then
+      sigmaT = ZERO
+    else
+      sigmaT = self % xsData % getTrackMatXS(p, p % matIdx())
+    end if
```
ChasingNeutrons, line 65: I'm amazed we hadn't seen this before. I'm curious whether this is the best place to put it rather than in the database? Putting it in the database might make for a tidier transportOperator, but might have some code duplication across different databases.

valeriaRaffuzzi: It's not a common problem because it can only happen if particles are born in void, and ST and DT already had checks to make sure that this is ok. 
I think adding the check to the tracking operator rather than the database is ok. In the database the only thing you could do is set all the xss to zero, but that might give you troubles somewhere else.

ChasingNeutrons (approved):

PR #153, opened 2025-02-06
Fixes to BEAVRS input

ChasingNeutrons:
BEAVRS could occasionally lose particles at the edge of the geometry due to the outermost cell being defined with the geometry bounding surface. There are likely ways to delete this class of problem altogether in future, but for now I've simplified the cell definitions in BEAVRS. I've also added in the HZP BEAVRS input which I understand has the D-bank of rods partially inserted.


valeriaRaffuzzi (approved):
Looks good! I had BEAVRS fail after a hundred something cycles and I imagine it was due to this... luckily it was an easy fix

PR #154, opened 2025-02-13
Energy depo

Esmae:
Implementation of mode 0 energy deposition tally as described in Tuominen, R., Valtavirta, V. and Leppänen, J., 2019. New energy deposition treatment in the Serpent 2 Monte Carlo transport code. Annals of Nuclear Energy, 129, pp.224-232. (Amongst other papers).

Works with macro or microResponse. Does not work with MG. 

Documentation is also updated.

### DIFF
--- a/NuclearData/xsPackages/neutronXsPackages_class.f90
+++ b/NuclearData/xsPackages/neutronXsPackages_class.f90
@@ -22,6 +22,7 @@ module neutronXsPackages_class
   !!   capture          -> sum of all reactions without secendary neutrons excluding fission [1/cm]
   !!   fission          -> total Fission MT=18 Cross-section [1/cm]
   !!   nuFission        -> total average neutron production Cross-section [1/cm]
+  !!   energyDepoZero   -> energy deposition, for mode zero. ejw89
   !!
   !!  Interface:
   !!    clean -> Set all XSs to 0.0
@@ -35,6 +36,7 @@ module neutronXsPackages_class
     real(defReal) :: capture          = ZERO
     real(defReal) :: fission          = ZERO
     real(defReal) :: nuFission        = ZERO
+    real(defReal) :: energyDepoZero   = ZERO
   contains
     procedure :: clean => clean_neutronMacroXSs
     procedure :: add   => add_neutronMacroXSs
@@ -54,6 +56,7 @@ module neutronXsPackages_class
   !!   capture          -> all reactions without secendary neutrons excluding fission [barn]
   !!   fission          -> total Fission MT=18 Cross-section [barn]
   !!   nuFission        -> total average neutron production Cross-section [barn]
+  !!   energyDepoZero   -> Energy Deposition using mode 0
   !!
   type, public :: neutronMicroXSs
     real(defReal) :: total            = ZERO
@@ -62,6 +65,7 @@ module neutronXsPackages_class
     real(defReal) :: capture          = ZERO
     real(defReal) :: fission          = ZERO
     real(defReal) :: nuFission        = ZERO
+    real(defReal) :: energyDepoZero   = ZERO
   contains
     procedure :: invert => invert_microXSs
   end type neutronMicroXSs
@@ -88,6 +92,7 @@ contains
     self % capture          = ZERO
     self % fission          = ZERO
     self % nuFission        = ZERO
+    self % energyDepoZero   = ZERO
 
   end subroutine clean_neutronMacroXSs
 
@@ -114,6 +119,7 @@ contains
     self % capture          = self % capture          + dens * micro % capture
     self % fission          = self % fission          + dens * micro % fission
     self % nuFission        = self % nuFission        + dens * micro % nuFission
+    self % energyDepoZero   = self % energyDepoZero   + dens * micro % energyDepoZero ! ejw89
 
   end subroutine add_neutronMacroXSs
 
@@ -152,6 +158,9 @@ contains
 
       case(macroAbsorbtion)
         xs = self % fission + self % capture
+
+      case(macroEnergyDepoZero)
+        xs = self % energyDepoZero
 
       case default
         xs = ZERO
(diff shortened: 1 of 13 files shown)


NuclearData/xsPackages/neutronXsPackages_class.f90
```diff
@@ -22,6 +22,7 @@ module neutronXsPackages_class
   !!   capture          -> sum of all reactions without secendary neutrons excluding fission [1/cm]
   !!   fission          -> total Fission MT=18 Cross-section [1/cm]
   !!   nuFission        -> total average neutron production Cross-section [1/cm]
+  !!   energyDepoZero   -> energy deposition, for mode zero.  [MeV/cm]
```
ChasingNeutrons, line 25: I would be inclined to rename this property. The reference to mode 0 is (I think) unique to Serpent, whereas the fission cross section * the Q value is commonly understood. Maybe qFission in analogy with nuFission would be more clear?

Tallies/TallyResponses/macroResponse_class.f90
```diff
@@ -119,12 +119,16 @@ function get(self, p, xsData) result(val)
...
     type(neutronMacroXSs)                 :: xss
     class(neutronMaterial), pointer       :: mat
+    character(100), parameter :: Here = 'get (macroResponse_class.f90)'        
 
     val = ZERO
 
     ! Return zero if particle is not neutron or if the particle is in void
     if (p % type /= P_NEUTRON) return
     if (p % matIdx() == VOID_MAT) return
+    if ((self % MT == macroEnergyDepoZero) .and. (p % isMG .eqv. .true.)) then
```
ChasingNeutrons, line 129: Perhaps for another PR (maybe an issue for now), but it is conceivable to have an MG equivalent. Basically this would mean giving the MG data one more input and it would be straightforward to compute qFission in a given material.

SharedModules/endfConstants.f90
```diff
@@ -110,6 +112,8 @@ module endfConstants
   integer(shortInt), parameter :: macroAllScatter = -20 ,&
                                   macroAbsorbtion = -21 ,&
                                   noInteraction   = -901
+  ! List of Macro Energy Deposition numbers for different energy deposition modes. Unique to SCONE
+  integer(shortInt), parameter :: macroEnergyDepoZero = -1100
```
ChasingNeutrons, line 116: I don't think there needs to be a number for mode 0 in particular at this stage. Rather, just energy deposition (or kappaFission) is fine, sitting with the other macros for consistency of style. Also, I might suggest making it -80 for consistency with Serpent. The micro would then be -301.
https://serpent.vtt.fi/mediawiki/index.php?title=ENDF_reaction_MT%27s_and_macroscopic_reaction_numbers

SharedModules/universalVariables.f90
```diff
@@ -80,7 +80,8 @@ module universalVariables
   real(defReal), parameter :: neutronMass = 939.56542194_defReal,  & ! Neutron mass in MeV (m*c^2)
                               lightSpeed  = 2.99792458e10_defReal, & ! Light speed in cm/s
                               kBoltzmann  = 1.380649e-23_defReal,  & ! Bolztmann constant in J/K
-                              energyPerFission = 200.0_defReal       ! MeV
+                              energyPerFission = 200.0_defReal,    &   ! MeV
```
ChasingNeutrons, line 83: I think this warrants some extra description. Is this supposed to be the estimated energy deposition of U235 for an LWR in particular? If so, I think its number is 202.27. Given that, what role does the scale factor below serve? I thought it wouldn't be necessary given Q235 is read from the ACE files?
I may be missing something though, I am not very knowledgeable on this.

Tallies/TallyResponses/Tests/macroResponse_test.f90
```diff
@@ -102,13 +111,15 @@ subroutine testGettingResponse(this)
     real(defReal), parameter :: TOL = 1.0E-9
 
     p % type = P_NEUTRON
+    p % isMG = .false.
```
ChasingNeutrons, line 114: Why did this have to be added? Perhaps I missed something. Is there a reason why energy deposition can't be included with MG? Perhaps for a second PR, but it would be nice to have a kappa value in the MG input that could be used.

Tallies/TallyResponses/microResponse_class.f90
```diff
@@ -150,13 +152,17 @@ function get(self, p, xsData) result(val)
...
     type(neutronMacroXSs)                 :: xss
-    character(100), parameter :: Here = 'get ( microResponse_class.f90)'
+    character(100), parameter :: Here = 'get (microResponse_class.f90)'
 
     val = ZERO
 
     ! Return zero if particle is not neutron or if the particle is in void
     if (p % type /= P_NEUTRON) return
     if (p % matIdx() == VOID_MAT) return
+    if ((self % MT == macroEnergyDepoZero) .and. (p % isMG .eqv. .true.)) then
```
ChasingNeutrons, line 162:
```suggestion
    if ((self % MT == macroEnergyDepoZero) .and. (p % isMG)) then
```

ChasingNeutrons (changes requested):
This all looks good to me. The only thing is the mentioned naming style. Something like kappaFission would be more appropriate I think as it is more generally understood.

SharedModules/universalVariables.f90
Mikolaj-A-Kowalski: Historical context: 
The `energyPerFission` line was added ad hoc as a 'by yesterday' way to have 'some' heat deposition estimate for power normalisation. (speaking of which I think that it was you @ChasingNeutrons who has added it :wink: )

But I have to say that I am not really happy that the heat deposition scaling fudge factor (which it is really) lives as a constant. It should really be a user defined parameter which should be specified at least per database basis. (maybe even per material? ).

This shouldn't be too bad to refactor I think. We just need to add a entry for database dictionary to read it and fudge the heating number in the database? But I didn't look to carefully to see if there are  any problems that could arise, so I might have missed something major.

PR #159, opened 2025-04-30
Ray plot

ChasingNeutrons:
I've implemented the ray trace plotting corresponding to that which was described recently in the Ridley and Nease papers at SNA+MC 2024.

The mains changes were adding normals to all surfaces, adding access procedures for normals to cells and universes, as well as the associated ray tracing subroutines which look slightly different from the usual ones.


valeriaRaffuzzi:
Addressed by #200.

PR #162, opened 2025-05-09
Add backtrace to error message for key errors

Mikolaj-A-Kowalski:
Fixes part of the usual user experience when working with the dictionary input. 
Now the dictionaries keep track of their address (path?) in the nested dictionary structure and on the key error when requesting the elements of dictionary the location is included in the error message. e.g.:
```
 <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
 Fatal has occurred in:
 getInt (dictionary_class.f90)                                                                       

 Because:
 In dictionary [/activeTally/flux/map] Entry under keyword N is not an 
 integer
 <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
```

The things to pay attention in the review are: 
- Some error messages that were missed and not updated 
- Potentially some use sequence of adding/removing/moving dictionaries that is not covered in the test. I fear I might have missed something and would appreciate 2nd pair of eyes 
- We may use this opportunity to reword/cleanup dictionary error messages

### DIFF
--- a/DataStructures/dictionary_class.f90
+++ b/DataStructures/dictionary_class.f90
@@ -121,6 +121,10 @@ module dictionary_class
     character(nameLen),dimension(:),allocatable  :: keywords
     type(dictContent),dimension(:), allocatable  :: entries
 
+    ! Meta data
+    ! TODO: public is bad! Get a getter!
+    character(:), allocatable, public :: address
+
     ! Dictionary state information
     integer(shortInt)                            :: maxSize = 0         ! Maximum size of a dictionary
     integer(shortInt)                            :: dictLen = 0         ! Current size of the dictionary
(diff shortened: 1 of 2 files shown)


DataStructures/dictionary_class.f90
```diff
@@ -549,7 +567,9 @@ subroutine getInt_new(self,value,keyword)
         value = self % entries(idx) % int0_alloc
 
       case default
-        call fatalError(Here,'Entry under keyword ' // keyword // ' is not an integer')
+        print *, self % address // keyword
```
Mikolaj-A-Kowalski, line 570:
```suggestion
```
What i**** made this PR! :angry: 
Leaving debug `printf`s!? Disgraceful!!!1

Mikolaj-A-Kowalski:
Partially addressing #155

valeriaRaffuzzi (commented):
This looks good to me and the test makes sense.. I don't think there's too many way to modify a dictionary that can affect its path!
I just don't get why deleting the path after a dictionary has been copied, especially if you're not deleting its nested dictionaries.

ChasingNeutrons (approved):

Issue #157, opened 2025-04-09
Problem reading thermal scattering data for h-poly.40t


Suark94:
I have a scone model that uses polyethylene and I try to read the thermal scattering data for h-poly.40 from ENDF80SaB2. When I try to do so, I get the error message below, pointing to SCONE/NuclearData/DataDecks/ACE/aceSabCard_class.f90:141 . For other materials, thermal scattering data can be read without any problems.

Error message:
<><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
 Fatal has occurred in:
 readIntArray (aceSabCard_class.f90)

 Because:
 Tried to read reals into integer array
 <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>

Mikolaj-A-Kowalski:
We are taking about this [ENDF80SaB2](https://nucleardata.lanl.gov/ace/endf80sab2/), right? 

The error itself means that the reader got confused about the locations in the XSS array and there are real numbers where one expects integers.

PR #163, opened 2025-06-05
Fix bug in building thermalScatterElastic_class.f90

valeriaRaffuzzi:
Two errors in one line! That was on me...

The mu distributions should be real rather than integers, and the size of the array to be read was also wrong.

My guess is that this didn't come up sooner because water and graphite (most commonly used) don't have this angular representation, while it showed up in poly.

This solves issue #157.


valeriaRaffuzzi:
> Would it be worth making an integration test for this?

There are some integration tests for Sab, but this was probably an edge case not captured.. Could be added as an issue for now.

PR #165, opened 2025-06-24
Rename densityResponse

ChasingNeutrons:
Maybe people will disagree with this...

I thought it might be appropriate to rename densityResponse to become invSpeedResponse. This is because, to me, it feels more natural to name the response what it is, rather than what it results in the tally being. Maybe we aren't fully consistent with this, e.g., fluxResponse.

I feel invSpeed is a bit more clear to someone reading rather than density, which is a bit more ambiguous in transport.

Finally, tallying 1/v weighted by flux is something one might do when computing kinetic parameters, so the response is a bit more general than producing a density.

If you disagree, let's fight!


valeriaRaffuzzi (approved):
Fine by me!

PR #166, opened 2025-06-28
K-implicit (n,2n) fix

ChasingNeutrons:
Previously k implicit only had a few multiplicative scattering channels it considered. This seemed to work well mostly, but, for JEFF with Be, this could cause significant keff bias. Hence I have added in MTs 875-891 which are discrete and continuous (n,2n) levels. These have also been added to endfConstants. I added a few others which we already had in endfConstants too, but in this case MTs 875-891 were the culprits.

### DIFF
--- a/Tallies/TallyClerks/keffImplicitClerk_class.f90
+++ b/Tallies/TallyClerks/keffImplicitClerk_class.f90
@@ -219,7 +219,7 @@ contains
     ! Select analog score
     ! Assumes N_XNs are by implicit weight change
     select case(MT)
-      case(N_2Nd, N_2N, N_2Na, N_2N2a, N_2Np, N_3Np, N_2N0:N_2Ncont)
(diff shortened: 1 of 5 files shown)


Tallies/TallyClerks/keffImplicitClerk_class.f90
```diff
@@ -219,9 +219,9 @@ subroutine reportOutColl(self, p, MT, muL, xsData, mem)
     ! Select analog score
     ! Assumes N_XNs are by implicit weight change
     select case(MT)
-      case(N_2N)
+      case(N_2Nd, N_2N, N_2Na, N_2N2a, N_2Np, N_3Np, N_2N0:N_2Ncont)
```
ChasingNeutrons, line 222:
```suggestion
      case(N_2Nd, N_2N, N_2Na, N_2N2a, N_2Np, N_2N0:N_2Ncont)
```

PR #167, opened 2025-07-09
Point-volume calculation and LDR benchmark

ChasingNeutrons:
I made the LDR Lite benchmark: https://serpent.vtt.fi/kraken/images/3/3d/LDR-PUB-VTT-100002-R2.pdf

This appears to agree well with Serpent.

When debugging it, I briefly convinced myself that the rayVolPhysicsPackage doesn't work. I think it does, but meanwhile I wrote a physics package that does the same by point sampling. Do we want this? If not, I can separate this out.


PhysicsPackages/pointVolPhysicsPackage_class.f90
```diff
@@ -0,0 +1,347 @@
...
+          if (matIdx /= OUTSIDE_MAT) exit rejection
+
+          i = i + 1
+          if (i > 1000) then
+            call fatalError(Here, 'Infinite loop when searching for a point in the geometry.')
+          end if
+        end do rejection
+    
+        ! Found something    
+        if (matIdx /= VOID_MAT) then
```
valeriaRaffuzzi: Why not calculate the volume of void too?

valeriaRaffuzzi (changes requested):
This is all very good, I just have the following comments:
- This PR include the lattice offset code; it might have to wait until lattice offset is merged
- It would be great if the volume calculation could be printed to output rather than to screen alone
- I would add VOID as a material for volume calculation purposes, I don't see why not
- MPI support would also be good

Sorry for the long time to review!

ChasingNeutrons:
In hindsight, I think the point volume calculation is unnecessary, and LDR is covered in another PR.

PR #171, opened 2025-08-01
Discontinuous field

ChasingNeutrons:
This adds a field which supports distance calculations to where the value of the field changes. This is hoped to support upcoming multiphysics features alongside both surface and delta tracking.

I have implemented an abstract class plus a derived type which allows a lattice-like spatial structure. Hopefully it should be easy to extend to more general fields.

In an upcoming PR I hope to add checks for this field for both temperature and density, allowing geometryStd to account for it during move.

Please let me know if this abstraction makes sense or doesn't.


Geometry/Fields/ScalarFields/PieceConstantFields/cartesianField_class.f90
```diff
@@ -0,0 +1,398 @@
...
+  public :: cartesianField_TptrCast
+
+  integer(shortInt), parameter :: ALL_MATS = -1
+
+  !!
+  !! Piecewise constant field constructed from a lattice-like grid. 
+  !! Values of the field are piecewise constant.
+  !!
+  !! Similar to a Cartesian lattice. Centre is placed at origin.
+  !! Can include materials. If applying the values uniformly to all materials, can use
```
valeriaRaffuzzi: It's unclear what 'materials' is: is it like a filter in this context?

Geometry/Fields/ScalarFields/PieceConstantFields/cartesianField_class.f90
```diff
@@ -0,0 +1,398 @@
...
+    do i = 1, size(mats)
+
+      call dict % get(temp, mats(i))
+
+      ! Flip array up-down for more natural input
+      ! Reshape into rank 2 array
+      tempMap = reshape(temp, [self % sizeN(1), self % sizeN(2) * self % sizeN(3)])
+      N = size(tempMap, 2)
+      do j = 1, N/2
+        call swap(tempMap(:,j), tempMap(:,N - j + 1))
```
valeriaRaffuzzi: It's not quite clear why you want to swap the array. Also, couldn't you make val 3D for better interpretability?

Geometry/Fields/ScalarFields/PieceConstantFields/cartesianField_class.f90
```diff
@@ -0,0 +1,398 @@
...
+  !!
+  !! Get distance to the next element of the field at the co-ordinate point and direction
+  !!
+  !! See pieceConstantField for details
+  !!
+  function distance(self, coords) result(d)
+    class(cartesianField), intent(in) :: self
+    class(coordList), intent(in)      :: coords
+    real(defReal)                     :: d
+    real(defReal)                     :: test_d
```
valeriaRaffuzzi: Could be in the same line

valeriaRaffuzzi (commented):
It all looks good to me, the main thing is that there is massive code duplication from latUniverse and it would be nice to avoid it. On the other hand these are not quite generic procedures and might never be used in a third location. Let's discuss it!

PR #172, opened 2025-08-22
Fix bug in S(a,b) coherent elastic scattering angle sampling

### LINKED ISSUE #169: Eigenvalue deviation from Serpent with thermal scattering
I tried to model the AKR zero-power reactor (https://tu-dresden.de/ing/maschinenwesen/ifvu/wket/ausbildungskernreaktor-akr-2/anlagenbeschreibung?set_language=en) and noticed that when thermal scattering was included in the materials, the eigenvalue obtained in SCONE deviated from the one in Serpent by about 2000 pcm. Without thermal scattering both values were the same. 

I then defined a substitute problem by only analysing a pin cell with fuel (UO2) and graphite as moderator (C-12). I did not notice any disagreement between SCONE and Serpent for regular pin cell dimensions. However, when I increased the pin cell to the dimensions of the AKR reactor (1m pitch, 25 cm pin diameter), I found that there was a deviation of around 500 pcm between SCONE (1.34716 +/- 0.00010) and Serpent (1.34198 +/- 0.00010) when using the thermal scattering file grph.40. Without thermal scattering k-eff agreed. 

I investigated a few variations of that pin cell problem (same dimensions). Changing boundary conditions from reflective to vaccuum, or replacing graphite with light water or polyethylene did not cause any issues when thermal scattering was activated, and the eigenvalues in SCONE and Serpent agreed with each other.

I used cross-section data at room temperature from the ENDF/B-8 library (https://nucleardata.lanl.gov/ace/lib80x/). The energy spectrum of the pin cell problem looks as follows:

Image

valeriaRaffuzzi:
This fixes #169 : there was an error in the coherent elastic scattering angle sampling. The CDF to sample the Bragg edge the neutron scatters with was built incorrectly.

This only came up when trying running a model with a lot of graphite; it wasn't noticeable when smaller models were simulated.

Now SCONE agrees perfectly with Serpent.

### DIFF
--- a/NuclearData/Reactions/thermalScattReactionCE/thermalScatterElastic_class.f90
+++ b/NuclearData/Reactions/thermalScattReactionCE/thermalScatterElastic_class.f90
@@ -234,26 +234,24 @@ contains
 
       ! Get energy indexes
       if (E_in > self % eIn(size(self % eIn))) then
-        l1 = size(self % eIn) - 1
+        l2 = size(self % eIn)
       else
-        l1 = binarySearch(self % eIn, E_in)
+        l2 = binarySearch(self % eIn, E_in)
       end if
 
-      l2 = l1 + 1
-
       ! Sample a Bragg edge at a smaller energy than E_in
-      prob = self % pValues(1:l2)
-      prob(1) = ZERO
-      r = rand % get() * prob(l2)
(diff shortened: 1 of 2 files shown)


ChasingNeutrons (approved):
Thanks for finding this, great stuff!

PR #173, opened 2025-08-24
Temperature and density fields

ChasingNeutrons:
Using the pieceConstantField, I have added the capability to super-impose temperature and density-scaling fields on the geometry. Both surface and delta tracking agree with each other for temperature and density changes, and they agree with reference calculations without imposed fields.

The distance calculations are done inside move. They work well for the most part, with one uncomfortable bit: fuzziness needed to be added to ensure that distance to the field does not take precedence over distance to the boundary. If there are any better suggestions for this it would be great.

This necessitated make the initMajorant function part of the nuclearDatabase, allowing it to be called in the physicsPackages which can receive the relevant field information from geometry.

I have updated the docs and added an integration test to geometry for obtaining values from the fields and distance calculations.


ParticleObjects/particle_class.f90
```diff
@@ -100,6 +100,10 @@ module particle_class
     real(defReal)              :: w         ! Particle Weight
     real(defReal)              :: time      ! Particle time point
 
+    ! Information passed from geometry
+    real(defReal)              :: T = -INF   ! Local temperature
```
valeriaRaffuzzi: You could use a consistent undefined value in between places: it is -INF here but, for example, -ONE in fissionSource_class.f90 (line 227)

NuclearData/nuclearDatabase_inter.f90
```diff
@@ -90,12 +91,36 @@ subroutine activate(self, activeMat, silent)
...
+    !! Optionally can scale the density. scaleDensity should be
+    !! the relative (to input) density of the highest density material.
+    !! This is most naturally used with a super-imposed density field.
+    !! As for temperature, this is conservative and could be improved with
+    !! a material-wise density scaling factor.
+    !!
+    subroutine initMajorant(self, loud, maxTemp, scaleDensity)
+      import :: nuclearDatabase, defBool, defReal
+      class(nuclearDatabase), intent(inout) :: self
+      logical(defBool), intent(in)          :: loud
```
valeriaRaffuzzi: Should `loud` be optional and default to false in implementation?

Geometry/Fields/ScalarFields/PieceConstantFields/cartesianField_class.f90
```diff
@@ -208,25 +209,45 @@ function at(self, coords) result(val)
...
-    if (self % matIdxs(1) /= ALL_MATS) then
-      idx = findloc(self % matIdxs, coords % matIdx, 1)
-      localID = localID + (idx - 1) * product(self % sizeN)
+    ! Ensure material is present
+    if (any(self % matIdxs == coords % matIdx)) then
+      idx0 = findloc(self % matIdxs, coords % matIdx, 1)
+      idx = idx + (idx0 - 1) * product(self % sizeN)
+    else if (self % matIdxs(1) == ALL_MATS) then
+      idx = idx
+    else
```
valeriaRaffuzzi:
```suggestion
    else if (self % matIdxs(1) /= ALL_MATS) then
```

NuclearData/ceNeutronData/ceNeutronCache_mod.f90
```diff
@@ -93,12 +101,14 @@ module ceNeutronCache_mod
   !! Structure that contains a cross section value
   !!
   !! Public Members:
-  !!   E  -> energy of the cross section
-  !!   xs -> value of the cross section
+  !!   E   -> energy of the cross section
+  !!   xs  -> value of the cross section
+  !!   T   -> temperature of the cross section
+  !!   rho -> density scaling of the cross section
```
valeriaRaffuzzi: Should T and rho have been added to cacheSingleXS?

NuclearData/ceNeutronData/ceNeutronDatabase_inter.f90
```diff
@@ -64,10 +64,10 @@ module ceNeutronDatabase_inter
 
     ! Procedures implemented by a specific CE Neutron Database
     procedure(updateMajorantXS), deferred     :: updateMajorantXS
-    procedure(updateTotalMatXS), deferred     :: updateTrackMatXS
+    procedure(updateTrackMatXS), deferred     :: updateTrackMatXS
```
valeriaRaffuzzi: How could we have survived with this massive typo?

Geometry/geometryStd_class.f90
```diff
@@ -202,24 +237,42 @@ subroutine move_noCache(self, coords, maxDist, event)
...
+    end if
+
+    if (maxDist < dist .and. maxDist < fieldDist) then ! Moves within cell
       call coords % moveLocal(maxDist, coords % nesting)
       event = COLL_EV
       maxDist = maxDist ! Left for explicitness. Compiler will not stand it anyway
-
+    
+    ! This check is really awful - can we do something better?
+    else if (fieldDist < dist .and. abs(fieldDist - dist) > 10*NUDGE) then ! Stays within the same cell, but crosses field boundary
```
valeriaRaffuzzi: Can't you just do 
else if (maxDist < dist .and. maxDist >= fieldDist)?
in this case you're not crossing any geometrical boundary but a field boundary.

valeriaRaffuzzi (changes requested):
That was a lot!!!! The nuclear data bit seemed to make sense but I have to admit it will be good to give it another look post revisions.

There are a couple of bigger issues that I wrote in specific comments, but in general it looks good! It was a massive effort so well done.

NuclearData/nuclearDatabase_inter.f90
```diff
@@ -90,12 +91,36 @@ subroutine activate(self, activeMat, silent)
...
+    !! Optionally can account for the temperature due to super-imposed
+    !! temperature fields. This should receive the maximum temperature in
+    !! the system. This is conservative and can be improved by creating a
+    !! material-wise maximum temperature input.
+    !!
+    !! Optionally can scale the density. scaleDensity should be
+    !! the relative (to input) density of the highest density material.
+    !! This is most naturally used with a super-imposed density field.
+    !! As for temperature, this is conservative and could be improved with
+    !! a material-wise density scaling factor.
```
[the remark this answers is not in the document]
ChasingNeutrons: I meant that the density approach is conservative. I have added some text to clarify that I am referring to temperature (above) and density here.

Tallies/TallyMaps/Maps1D/fieldMap_class.f90
```diff
@@ -0,0 +1,182 @@
...
+    integer(shortInt)                :: i
+
+    ! Name the array
+    name = trim(self % getAxisName()) // 'Bins'
+
+    call out % startArray(name, [1, self % Nbins])
+
+    ! Print field indexes
+    do i = 1, self % Nbins
+      call out % addValue(numToChar(i))
```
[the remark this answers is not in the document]
ChasingNeutrons: Maybe this can be more sophisticated in future, but I think there should be some index information available directly for postprocessing. Do you suggest not having it?

docs/User Manual.rst
```diff
@@ -814,6 +822,52 @@ Example: ::
...
+  - pitch: (x y z) array with the x, y and z field pitches. In a 2D field,
+    the value entered in the third dimension is not used. [cm]
+  - origin (*optional*, default = (0.0 0.0 0.0)): (x y z) array with the
+    origin of the field. [cm]
+  - materials: list of material names, corresponding to materials in nuclearData.
+    Optionally, ``all`` can be used, applying the values of the field to all materials.
+  - names of each material: a map, named after every material present in the materials list. 
+    The entries of the map are the values that the field takes in that material in that
+    element of the field. The order is: increasing x, increasing y and then increasing z.
+  - default: the value taken by the field when a point is either outside of the field or
```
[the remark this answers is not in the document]
ChasingNeutrons: Oh, I think I understand better now. You mean for each material specified in the field definition. This does make sense; the default option was made for anything that doesn't correspond to what's in the map (either in space or material). There is maybe quite a multiplicity of options to consider...

Geometry/Fields/ScalarFields/PieceConstantFields/pieceConstantFieldFactory_func.f90
```diff
@@ -0,0 +1,58 @@
+!!
```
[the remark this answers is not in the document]
ChasingNeutrons: I don't know quite whether this would work without breaking other things. For example, there is the addField which is called in geometry_registry. How exactly do we use it? It might be a neater way of including temperature and density fields than what I currently do, but I think it needs some explaining.

NuclearData/ceNeutronData/ceNeutronMaterial_class.f90
```diff
@@ -446,13 +478,15 @@ end subroutine sampleNuclide
   !!   fatalError if E is out-of-bounds of the present data
   !!   Returns nucIdx <= if material is not fissile
   !!
-  function sampleFission(self, E, rand) result(nucIdx)
+  function sampleFission(self, E, temp, rho, rand) result(nucIdx)
```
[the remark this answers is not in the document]
ChasingNeutrons: Maybe best just to leave them for now in case we do need them in future? sampleFission, at least, is used in fissionSource.

ChasingNeutrons:
Made some changes to this:

1. I removed the fieldMap. It is unnecessary just for the moment and I don't think I had a good idea of how it would be used just yet. It might require a few more changes to fields to make it work well.
2. I made the temperature and density fields much more like other fields, i.e., they are kept in the geometry registry, not in the geometry. This was just my ignorance before. I think this is much neater. I made some small changes to the registry to improve performance for these fields, because the character map can have a bit of overhead.

Otherwise, I think I have addressed most of the comments.

PR #180, opened 2025-11-17
Event tally

ChasingNeutrons:
This adds a non-standard tally which records individual events. This can be used for, e.g., detector modelling. It does not use scoreMemory, but instead keeps a record of all events and periodically outputs them to a file.

This could be improved with a change to the tally interface or the particle to send information about the colliding nuclide to the tally.

Future complementary developments might include more analogue features in the collisions, e.g., more physical multiplicity models and fission neutron correlations.

If this looks good I will write some tests.

### DIFF
--- a/Tallies/TallyClerks/eventClerk_class.f90
+++ b/Tallies/TallyClerks/eventClerk_class.f90
@@ -140,8 +140,7 @@ module eventClerk_class
     ! Output procedures
     procedure           :: display
     procedure           :: print
-    procedure, private  :: outputScoresText
-    procedure, private  :: outputScoresBinary
+    procedure, private  :: outputScores
 
   end type eventClerk
 
@@ -313,11 +312,7 @@ contains
         self % lastEntry(id) = 0
       else
 
-        if (self % binary) then
-          call self % outputScoresBinary(id)
-        else
-          call self % outputScoresText(id)
-        end if
+        call self % outputScores(id)
       
         ! Increment the total number of events
         self % totalEvents = self % totalEvents + self % lastEntry(id)
@@ -358,13 +353,7 @@ contains
     ! Make sure that score arrays have been flushed already.
     ! If not, write the remainder to output.
     do i = 1, size(self % lastEntry)
-      if (self % lastEntry(i) > 0) then
-        if (self % binary) then
-          call self % outputScoresBinary(i)
-        else
-          call self % outputScoresText(i)
-        end if
-      end if
+      if (self % lastEntry(i) > 0) call self % outputScores(i)
     end do
 
     ! Write out any map that was used to the usual output file,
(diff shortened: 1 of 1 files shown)


InputFiles/Benchmarks/BEAVRS/BEAVRS_HZP
```diff
@@ -232,6 +232,16 @@ geometry {
     NS3 { type simpleCell; id 12; surfaces (-4 5 -10 11); filltype mat; material SS304;}
     NS4 { type simpleCell; id 13; surfaces (-4 5 10 -11); filltype mat; material SS304;}
     
+<<<<<<< HEAD
```
valeriaRaffuzzi: Damn BEAVRS!!! It's everywhere

Tallies/TallyClerks/eventClerk_class.f90
```diff
@@ -0,0 +1,421 @@
...
+  subroutine outputScores(self, id)
+    class(eventClerk), intent(in) :: self
+    integer(shortInt), intent(in) :: id
+    integer(shortInt)             :: i, u, ios
+
+    associate(ev => self % scores(:,id))
+
+
+      ! Open the file and append the event info
+      open(newunit=u, file=trim(self % outputFile), status='unknown', &
```
valeriaRaffuzzi: Here you are either creating the file or opening it, right? Is `append` happy with either?

docs/User Manual.rst
```diff
@@ -1102,6 +1102,32 @@ Example: ::
...
+  underwent. Maps are treated as filters, only returning events which fall into the map.
+  Events can be post-processed, e.g., for alpha calculations.
+
+  - file: path to the file which records events.
+  - maxScale (*optional*): a real scaling factor on the maximum number of events before 
+    recording ceases. By default, the maximum number of events is 10M. This is a scaling
+    factor rather than an integer due to constraints on the dictionary.
+  - freq (*optional*): an integer determining how often events are written to the file.
+    More often will incur more parallel overhead, less often will incur a larger memory
+    footprint. Set to 500k by default.
```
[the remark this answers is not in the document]
ChasingNeutrons: It's about 960MB with these settings. I have now included it in the docs.

Tallies/TallyClerks/eventClerk_class.f90
```diff
@@ -0,0 +1,421 @@
...
+    real(defFlt), dimension(3) :: dir
+    real(defFlt)               :: Eincident
+    real(defFlt)               :: time
+    real(defFlt)               :: w
+    integer(shortInt)          :: brood
+    
+    ! Collision info
+    real(defFlt)                :: Edeposit
+    integer(shortInt)           :: MT
+    !integer(shortInt)           :: ZAID ! Don't yet have available data in interface
```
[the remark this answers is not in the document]
ChasingNeutrons: The question here is maybe deciding what to do for MG. Let me make an issue and we can add that easily enough if needs be.

Tallies/TallyClerks/eventClerk_class.f90
```diff
@@ -0,0 +1,470 @@
...
+      close(u)
+
+    end associate
+
+  end subroutine outputScoresText
+  
+  !!
+  !! Send events in a thread's bank to a binary output file
+  !!
+  subroutine outputScoresBinary(self, id)
```
valeriaRaffuzzi: I don't want to be annoying but this could have been done more compactly: the only difference between procedures is access. It could have been one function called outputScores (noting that form = 'unformatted' is default):

if (self % binary):
   accessType = 'stream'
else
   accessType = 'sequential'
end if

open(newunit=u, file=trim(self % outputFile), status='unknown', &
           access=accessType, position='append', &
           action='write', iostat=ios)

Issue #187, opened 2026-01-12
MT 4 tallies giving the same result as MT -4


bonnettheophile:
Following ENDF/Serpent wiki, MT 4 should correspond to (n,n) inelastic scattering only, whereas using microResponse or macroResponse with MT 4 in the current implementation gives the total inelastic scattering, including (n,2n), etc.

valeriaRaffuzzi:
Sorted with PR #189.

Issue #123, opened 2024-04-18
Energy printing in MGxsClerk


valeriaRaffuzzi:
Right now results from MGxsClerk are printed fast -> thermal in energy, but the energy map is not which could be confusing. 

I would flip the order of the printed energy map for this clerk, and write in the user manual which order to expect.

PR #189, opened 2026-01-15
Tally improvements

valeriaRaffuzzi:
This PR addresses two issues:

- it flips the order of the energy grid being printed out by the mgXsClerk, which is now fast -> thermal, consistently with the order of the cross sections and scattering matrices (issue #123)
- it conforms the MT numbers that one can request with the tallies with the ENDF definitions:
       - MT = 3 is total non-elastic (= inelastic scattering + disappearance + fission)
       - macroCapture was renamed to macroDisappearance, which is the equivalent of MT = 101
       - MT = 4 is the sum of MT 50 -> 91. MT 4 is not normally present in the ACE files (although sometimes weirdly it's present between the capture reactions), so now it is artificially constructed inside aceNeutronNuclide. Importantly, MT = 4 is not the same as mainData(IESCATTER_XS,:), which also contains (n,xn) and other channels. Now this is all consistent, addressing issue #187.

### DIFF
--- a/docs/Input Manual.rst
+++ b/docs/Input Manual.rst
@@ -1329,8 +1329,8 @@ Example: ::
       collision_estimator { type collisionClerk; response (flux); flux { type fluxResponse; } }
       }
 
-* invSpeedResponse: used to calculate flux-weighted inverse speed or the particle density, i.e., the response function is 
-  the inverse of the particle speed in [cm/s]
+* invSpeedResponse: used to calculate flux-weighted inverse speed or the particle density, i.e., the 
+  response function is the inverse of the particle speed in [cm/s]
 
 Example: ::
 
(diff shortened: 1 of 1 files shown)


docs/Input Manual.rst
```diff
@@ -1342,7 +1342,7 @@ Example: ::
 
   - MT: MT number of the desired reaction. The options are: -1 (total), -2 (disappearance),
     -3 (elastic scattering), -4 (total inelastic scattering), -6 (fission), -7 nu*fission),
-    -20 (total scattering), -21 (absorption).
+    -20 (total scattering), -21 (absorption), -22 (total non elastic).
```
ChasingNeutrons: Maybe, since it is the manual, it is better to be more explicit and saying 'total all reactions except elastic' to avoid the ambiguity with inelastic.

Issue #195, opened 2026-02-14
Unparallelising combing


ChasingNeutrons:
I owe both @valeriaRaffuzzi and @martin220199 an apology: the parallel comb I implemented is actually not very good. This is because the inner while loop makes it N^2, which really dominates runtime. I think, at least for now, it's much better to just be serial! I'll revert this shortly, for both neutrons and precursors.

PR #196, opened 2026-02-17
Alpha solver

ChasingNeutrons:
This contains the modifications and physicsPackage needed to solve alpha eigenvalue problems. This adds alpha and a stabilisation factor, eta, to the particle, and obtains the appropriate alpha cross section during transport. I also included Cullen's example problems. I also added an implicit removal time tally.

While doing this, I made some small tweaks to kinetic to improve the performance of combing as mentioned in issue #195, added optional time dependence to materialSource (which we should probably make more general in future) and fixed a bug in the kineticPhysicsPackage which prevented the time dependence of sources from being properly accounted for.

Finally, I added an optional to keffImplicitClerk to tally only prompt or delayed neutron production, rather than the total that we previously had.

### DIFF
--- a/ParticleObjects/particle_class.f90
+++ b/ParticleObjects/particle_class.f90
@@ -465,9 +465,6 @@ contains
   !!   Further there is currently no good solution for MG neutrons. Their speed
   !!   is arbitrarily set to 1.
   !!
-  !! Does not provide errors if a dubious particle type is used. This is to allow
-  !! the function and related functions to be pure.
-  !!
   !! Args:
   !!   None
   !!


Tallies/TallyClerks/keffImplicitClerk_class.f90
```diff
@@ -109,6 +121,13 @@ subroutine init(self, dict, name)
     ! Handle virtual collisions
     call dict % getOrDefault(self % handleVirtual,'handleVirtual', .true.)
 
+    ! Which fission production to tally? Total, prompt, or delayed?
+    call dict % getOrDefault(self % setting,'setting', 1)
```
valeriaRaffuzzi: I would use TOTAL instead of 1 here.

NuclearData/ceNeutronData/aceDatabase/aceNeutronNuclide_class.f90
```diff
@@ -110,6 +111,7 @@ module aceNeutronNuclide_class
     character(nameLen)                          :: ZAID    = ''
     real(defReal), dimension(:), allocatable    :: eGrid
     real(defReal), dimension(:,:), allocatable  :: mainData
+    real(defReal), dimension(:,:), allocatable  :: promptNuFiss
```
valeriaRaffuzzi: This line is probably a leftover

Issue #207, opened 2026-04-08
MT-5 sampling neutrons below minimum energy


ChasingNeutrons:
I have observed _extremely rarely_ MT-5 reactions on U-235 producing particles at energies below 1E-11, causing a failure in grid searches. This is likely a data problem. This problem can likely be avoided by setting an energy floor in one of the scattering reactions and also in scatteringKernels.

PR #208, opened 2026-04-08
MT5 fix

ChasingNeutrons:
This addresses #207 by putting a floor on the neutron energy sampled from neutronScatter. I also added in two new variables: MINIMUM_ENERGY and MAXIMUM_ENERGY. We don't actually use 1E-11 and 20.0 all that often in the code, but there are a handful of places - maybe we should use them more? There might be some places I have forgotten to include these where they might be relevant, so let me know if you think of any.


valeriaRaffuzzi (approved):
Thanks for the fix, looks good! 
One thing to note is that the energy boundaries in the collision Operator are independent from those of the nucleardatabase. This is something we might want to unify at some point, if possible. But for now this is ok, since it just replaces a hardcoded parameter by a global variable.

PR #210, opened 2026-04-19
tiny-typo-fix

shimwell:
Just reading the README and noticed some tiny typos.


ChasingNeutrons (approved):
Thanks for the catch! The workshop is having its intended effect!

PR #215, opened 2026-05-01
Fix rare bug related to TMS

### LINKED ISSUE #214: Failure to sample inelastic scattering correctly
I have observed an extremely rare crash where inelastic scattering is sampled but the code fails to sample correctly from the angular distribution. Obtaining some diagnostic info, this seems to happen for high energies and _just_ outside of the energy grid for inelastic. This suggests some index/interpolation inconsistency with these cross sections. Admittedly, I have observed this on a branch I am modifying (but I don't have a strong reason to believe my modifications have affected this) and compiled with gfortran 13.3.0.

valeriaRaffuzzi:
This fixes #214: there was an inconsistency in the energy used to sample a reaction and sample the outgoing state of a particle after scattering when using TMS.

When using TMS, a relative energy is sampled in 'ceNeutronMaterial % sampleNuclide' for each nuclide. That energy is that saved in collDat % E and subsequently used to retrieve all cross sections. However, during scattering, p % E was used to sample the outgoing particle. 

This was now fixed in neutronCEstd and neutronCEimp.


ChasingNeutrons (approved):
Looks very good to me. Thank you for finding this sneaky bug which may have been doing some subtle damage!

PR #216, opened 2026-05-06
Fix MG majorant allocation bug

Esmae:
Earlier this year a change to the physics packages mean they can also allocate majorants. To match with CE, MG now needs a check to deallocate if required.


ChasingNeutrons:
Thanks for catching this. Strange it didn't rear its head before! I'll approve once tests are done.

valeriaRaffuzzi (approved):
Looks good, thanks Esmae!

PR #218, opened 2026-07-11
Dispersion and pebble universes

ChasingNeutrons:
I have added a universe which contains spherical inclusions, e.g., for TRISO-containing systems. It reads a list of sphere radii, origins, and fills, and populates the volume with them. Cell searches are accelerated by a grid (much like Serpent does, I think) which precomputes which cells are in a given grid cell. I also added in a spherical universe - equivalent to a pinUniverse in spherical co-ordinates.

As part of that work, I also refactored Cartesian grids in general. This universe was the third place in the code they were used, so I made a CartesianLattice class which those places (dispersionUni, latUni, and cartesianField) all now use. It sits in SharedModules, but perhaps it is more appropriate in geometry somewhere?

Below is an image generated using the ray tracing visualisation. With respect to performance, intriguingly this particular problem (a pebble with typical packing fraction, in a reflected boundary box) was faster with pure delta tracking than with our hybrid tracking default. This perhaps suggests I could optimize the distance calculation a bit more. Distance caching helps performance dramatically. Having it on shaved off a quarter of the runtime compared to without.
image
This universe was made with a script to generate particle positions. I should perhaps add this (or something like it) into cream, as well as an example input file.


nuclearkevin:
Very cool @ChasingNeutrons! Do you have a sense of how much of a performance improvement you get with a dispersed universe compared to a standard universe with many TRISO particles added to it (with cell sorting)? I've been trying to convince myself (and others) that its worthwhile to implement a custom universe type for dispersed fuels in OpenMC haha

ChasingNeutrons:
Hey @nuclearkevin, thanks for the interest! Let me run this case tonight and get back to you. 
One very brief investigation I was surprised by was that this structure seemed surprisingly sensitive to the nuclide densities I was using. I got some TRISO and surroundings densities from AI (first mistake...) and was a bit underwhelmed by the performance of hybrid delta tracking - pure delta tracking was way better! This was especially because Jaakko had previously reported that hybrid tracking worked well for exactly the same problem type here: https://serpent.vtt.fi/mtg/2022_Munich/Leppanen3.pdf
I then ran Serpent on the same geometry as for SCONE, but with the densities taken from here (with their f34 fuel type) (https://serpent.vtt.fi/mediawiki/index.php/HTTR_full_core_model) and found it to be way faster! In desperation, I replaced the densities with my hallucinated ones, however, and it slowed down quite a bit and indeed showed that delta tracking was better in that case. I haven't closed the loop and tried the Serpent densities in SCONE, but that's next on the list. It's definitely an intriguing geometry type...

ChasingNeutrons:
I ran a pure cell pebble, containing spheres of pure UO2 at 15.5% enriched (rather than TRISO, for simplicity), with packing fraction of 30%, giving about 48k particles. On my Intel Gold node using 40 threads this was estimated (I didn't run to completion) to take an hour to do 50 cycles with 20k particles per cycle when using pure delta tracking. With pure surface tracking (and cell neighbour lists), it was estimated at 8.5 hours.
The specialised universe times were 2 seconds and 8 seconds, respectively.
I don't know what you guys do at the moment, but I think a dedicated universe will definitely give significant performance benefits!

nuclearkevin:
Thanks for indulging my curiosity @ChasingNeutrons! Right now OpenMC doesn't apply any specific acceleration structure for distributed fuels like TRISO. Instead, we create a series of rectangular prism "search universes" in the Python API and assign each particle to one or more of those prisms. This helps, but creates a lot of extra cells/universes in the geometry and requires a good deal of user intuition to get reasonable performance (as reasonable as possible seeing as OpenMC's main branch uses pure surface tracking).

I've implemented (hybrid) delta tracking in OpenMC which has gotten us a decent improvement in FoM for TRISO. This (and Serpent) convinces me that a TRISO-specific universe is definitely the way to go to further improve performance. Now I just need to convince Patrick to let me spend some time on it. :)

ChasingNeutrons:
One thing I have done in making our acceleration structure was to size the number of grid cells in each dimension as the width in that dimension (given by the bounding box for all of the particles) divided by the average radius of the particles. It felt like a reasonable thing to do, but I haven't explored it. I would be interested to see if there are any more clever approaches. Good luck making your case to Patrick!

PR #221, opened 2026-07-23
Comminity and Contributing documents

Mikolaj-A-Kowalski:
Hi all!

Since the SCONE community might be getting more mature it may be a good idea to add some documentation about contributing and have a document that will describe the principles that should govern our community and the SCONE repository.

I have written a draft in this PR. However, it is basically my opinion/preference (rumblings?) put to text. It should not be taken as anything final. It is best to think of it as a starting point for discussion. Not only on the documents, but our community as a whole.

I kind of have a feeling that there is a lot of people working on the code, developing new functionality, but never pushing the changes upstream to the  main repository. I would kindly ask you all to contemplate why do we think it is that way? What are the blockers to contributing? What can we do to remove them?

Basically if the SCONE is not indifferent to you, please consider contributing to the discussion here.

`><))))*>`

PR #219, opened 2026-07-13
Governance

valeriaRaffuzzi:
Adding a SCONE Governance document to the docs. This details the expected behaviour of contributors and maintainers, and how decisions will be taken in the future.

PR #222, opened 2026-07-23
Revert "Governance (#219)"

Mikolaj-A-Kowalski:
To make place for the new version following public discussion in #221


valeriaRaffuzzi:
Thanks for this. I suggest approving this when PR #221 is accepted.

